PR #1, opened 2021-03-11
Cmake changes and version bump

tfalders:
Ports changes from https://github.com/ROCmSoftwarePlatform/hipBLAS/pull/295 to hipSOLVER and bumps hipSOLVER's version number.

Now that hipSOLVER is public, I would like to start opening PRs and getting code reviews from @jzuniga-amd and @cgmb as normal, rather than pushing straight to the repo.


library/src/CMakeLists.txt
```diff
@@ -61,14 +61,10 @@ if( NOT CUDA_FOUND )
     find_package( rocsolver REQUIRED CONFIG PATHS /opt/rocm /opt/rocm/rocsolver /usr/local/rocsolver )
   endif( )
 
-  target_compile_definitions( hipsolver PRIVATE __HIP_PLATFORM_HCC__ )
-
-  target_link_libraries( hipsolver PRIVATE roc::rocblas roc::rocsolver )
+  target_link_libraries( hipsolver PRIVATE roc::rocblas roc::rocsolver hip::host )
```
cgmb, line 64: This matches hipBLAS, but I think hipBLAS is probably not handling this correctly. These should probably be `PUBLIC`. Don't worry about it for now, though. The change you have made is minimal and safe. We can investigate and address that issue separately at a later date (it's relatively minor, but requires some investigation and testing).

jzuniga-amd (approved):
LGTM

cgmb (approved):
My comments are not matters of immediate concern. Just things to be aware of.

CMakeLists.txt
```diff
@@ -49,7 +49,7 @@ include( ROCMInstallTargets )
 include( ROCMPackageConfigHelpers )
 include( ROCMInstallSymlinks )
 
-set ( VERSION_STRING "0.1.0" )
+set ( VERSION_STRING "0.2.0" )
```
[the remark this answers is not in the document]
jzuniga-amd: Yeah, I think we should bump to version 1.0.0 soon. Maybe once all the wrappers are in, which should be by the end of the sprint. So, I see no problem changing to 1.0.0 in this PR.

library/src/CMakeLists.txt
tfalders: Yeah, at some point I may have to ask you to review the cmake system for hipSOLVER and hopefully clean it up like you've done for rocSOLVER.

PR #2, opened 2021-03-12
Updated README

tfalders:


cgmb (approved):
LGTM.

jzuniga-amd (approved):
LGTM

library/src/hcc_detail/hipsolver.cpp
```diff
@@ -90,8 +90,9 @@ hipsolverStatus_t rocblas2hip_status(rocblas_status_ error)
         return HIPSOLVER_STATUS_ALLOC_FAILED;
     case rocblas_status_internal_error:
         return HIPSOLVER_STATUS_INTERNAL_ERROR;
+    default:
+        throw "Unimplemented status";
     }
-    throw "Unimplemented status";
```
cgmb, line 94: I assume this was copied from hipBLAS, but FYI, throwing strings is discouraged by the [C++ Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#e14-use-purpose-designed-user-defined-types-as-exceptions-not-built-in-types). Aside from the reasons described there, the built-in termination handler also prints a more useful error message when receiving an uncaught exception derived from `std::exception` (at least on Ubuntu 18.04).

No need to change anything right now. This works fine. I just wanted to note that it's not ideal, so we don't copy bad practices when writing new code.

PR #4, opened 2021-03-25
Minor fixes

tfalders:
Improves exception handling and uses #pragma once across library headers as suggested by @cgmb.

Also removes HIPSOLVER_FILL_MODE_FULL, as this fill mode is missing from cuSOLVER. While this is a change to the public API, there are currently no functions that use hipsolverFillMode_t and the library is so early in development that I don't believe a deprecation process is required.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,14 @@
 # Change Log for hipSOLVER
 
-## [(Unreleased)]
+
+## [(Unreleased) hipSOLVER 1.0.0 for ROCm 4.3]
+### Added
+
+### Removed
+- Removed unused HIPSOLVER_FILL_MODE_FULL enum value.
+
+
+## [hipSOLVER 0.1.0 for ROCm 4.2]
 ### Added
 - Created hipSOLVER repository
 - Added functions
--- a/clients/benchmarks/CMakeLists.txt
+++ b/clients/benchmarks/CMakeLists.txt
@@ -40,6 +40,7 @@ target_include_directories( hipsolver-bench
 # External header includes included as system files
 target_include_directories( hipsolver-bench
   SYSTEM PRIVATE
+    $<BUILD_INTERFACE:${CBLAS_INCLUDE_DIRS}>
     $<BUILD_INTERFACE:${HIP_INCLUDE_DIRS}>
 )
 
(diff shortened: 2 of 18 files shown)


cgmb:
You're updating the throws, but I don't see any catch statements being updated... where do these exceptions end up getting caught?

tfalders:
> You're updating the throws, but I don't see any catch statements being updated... where do these exceptions end up getting caught?

Ah, I didn't add anything like that in. I'll go do that now.

cgmb:
Changing the error handling strategy for the API would be a very big change. That would require a discussion with @daineAMD to better understand the current error-handling strategy we copied from hipBLAS, and to coordinate a common strategy with hipBLAS for any changes we make. I think we might want to just separate that change from the rest of the PR.

Your other changes look good to me, though I don't know anything about `HIPSOLVER_FILL_MODE_FULL`, so I'd be trusting you and Juan on whether that's important or not. (It would need a changelog entry, though.)

tfalders:
> Changing the error handling strategy for the API would be a very big change. That would require a discussion with @daineAMD to better understand the current error-handling strategy we copied from hipBLAS, and to coordinate a common strategy with hipBLAS for any changes we make. I think we might want to separate that change from the rest of the PR and arrange a discussion.

@daineAMD indicated to me that he's okay with the change in this PR, but exception handling for the parameter conversion will be a bigger change and warrants a discussion.

jzuniga-amd (approved):

jzuniga-amd (approved):

cgmb (changes requested):
I have no fundamental objections, but we still need to have that discussion about the error handling.

There are three places where problems can come from: user code, rocSOLVER/cuSOLVER, and hipSOLVER itself. We need to clearly define our strategy for how we plan to handle errors stemming from each of those three origins.

I want to be sure we're all on the same page and that our code matches the strategy we've chosen. (There might not actually be any code changes required.)

tfalders:
> I have no fundamental objections, but we still need to have that discussion about the error handling.

Agreed, but any changes we wish to make on the error handling side I will probably add in a separate PR (along with bad args tests as requested by Juan). It could be a large change, depending on the strategy. I'd like to merge this PR soon, if possible.

cgmb (approved):
Sounds good. This API change is likely to be a part of any strategy we choose, so go ahead and merge if that helps with integrating your other work. We just should be sure to have the full details worked out before the release cut-off.

PR #5, opened 2021-03-25
Updated changelog for ROCm 4.2

tfalders:
Updates changelog for final release of ROCm 4.2. This is a minor change, so I will probably merge with only one approval from @jzuniga-amd.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Change Log for hipSOLVER
 
-## [(Unreleased)]
+## [hipSOLVER 0.1.0 for ROCm 4.2]
 ### Added
 - Created hipSOLVER repository
 - Added functions


jzuniga-amd (approved):
LGTM.
(This change has to be ported back to develop at some point. What we do in rocSOLVER is to prepare the changelog when we "release" from development to master at feature compete -the end of the development cycle- I suggest we do the same here for future releases to avoid having these "update changelog" commits cascading back...)

cgmb (approved):

tfalders:
> LGTM.
> (This change has to be ported back to develop at some point. What we do in rocSOLVER is to prepare the changelog when we "release" from development to master at feature compete -the end of the development cycle- I suggest we do the same here for future releases to avoid having these "update changelog" commits cascading back...)

Agreed. I forgot to do it this time, but in the future I'll try and keep these changes in the develop branch.

PR #10, opened 2021-04-12
Improve exception handling

tfalders:
When passing an invalid enum, functions will now return an error status rather than throwing an exception.


cgmb (approved):
The hipSOLVER wrapper is so thin that we could probably get away without this, but it seems like a cautious initial solution that we can improve upon later. I have no objections.

I do want to look into exactly what guarantees we get on the value of `hipsolverStatus_t(-1)` in C and C++, but if we're not depending on that in the library code, that can be an investigation for another time.

jzuniga-amd (approved):
LGTM

PR #11, opened 2021-04-12
Added getrs

tfalders:
Adds getrs to hipSOLVER

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 ## [(Unreleased) hipSOLVER 1.0.0 for ROCm 4.3]
 ### Added
 - Added functions
+  - getrs
+    - hipsolverSgetrs, hipsolverDgetrs, hipsolverCgetrs, hipsolverZgetrs
   - potrf
     - hipsolverSpotrf_bufferSize, hipsolverDpotrf_bufferSize, hipsolverCpotrf_bufferSize, hipsolverZpotrf_bufferSize
     - hipsolverSpotrf, hipsolverDpotrf, hipsolverCpotrf, hipsolverZpotrf
--- a/README.md
+++ b/README.md
@@ -104,5 +104,6 @@ hipsolverSgetrf(hipsolverHandle_t handle,
 | -------- | ------ | ------ | -------------- | -------------- |
 | hipsolverXgetrf_bufferSize | x | x | x | x |
 | hipsolverXgetrf | x | x | x | x |
+| hipsolverXgetrs | x | x | x | x |
 | hipsolverXpotrf_bufferSize | x | x | x | x |
 | hipsolverXpotrf | x | x | x | x |
--- a/clients/benchmarks/client.cpp
+++ b/clients/benchmarks/client.cpp
@@ -68,7 +68,7 @@ try
         ("function,f",
          value<std::string>(&function)->default_value("getrf"),
             "The LAPACK function to test.\n"
-            "                           Options are: getrf, potrf.\n"
+            "                           Options are: getrf, getrs, potrf.\n"
             "                           ")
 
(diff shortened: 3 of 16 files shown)


tfalders:
I appear to have run the bad_arg tests on the CUDA machine at a time when the tests weren't actually triggering (see 
https://github.com/ROCmSoftwarePlatform/hipSOLVER/pull/10/commits/ed3b239a45a295424d5277b4630b257c0527bc36). Running them properly is causing failures. The nullptr checks, in particular, seem to be causing all future tests to fail, so I've changed the bad_args tests to only check the invalid enum returns.

cgmb (approved):
LGTM.

I have no strong opinions on how the API should handle null pointers.

jzuniga-amd:
> I appear to have run the bad_arg tests on the CUDA machine at a time when the tests weren't actually triggering (see
> [ed3b239](https://github.com/ROCmSoftwarePlatform/hipSOLVER/commit/ed3b239a45a295424d5277b4630b257c0527bc36)). Running them properly is causing failures. The nullptr checks, in particular, seem to be causing all future tests to fail, so I've changed the bad_args tests to only check the invalid enum returns.

And so, what would actually happen if a user passes an invalid pointer?
The rocsolver backend will return rocblas_status_invalid_pointer, but how this would be translated to a hipsolver status? Is there a way to test this?
What about on the cusolver side?

tfalders:
> And so, what would actually happen if a user passes an invalid pointer?
> The rocsolver backend will return rocblas_status_invalid_pointer, but how this would be translated to a hipsolver status? Is there a way to test this?
> What about on the cusolver side?
> 
> I think this needs to be addressed not eliminated from the tests...

With the rocsolver backend, rocblas_status_invalid_pointer is translated to HIPSOLVER_STATUS_INVALID_VALUE and the tests pass. Passing a nullptr into the cuSOLVER backend does not appear to return an error - the returned status is HIPSOLVER_STATUS_SUCCESS and the tests fail. Worse, it seems to cause issues with future hip calls - I've gotten errors from hipFree and our memcheck calls after passing a nullptr.

I'm not sure there's any way to test this, since cuSOLVER is not reporting an error and it causes failures in all future tests.

tfalders:
I suppose I could put nullptr checks into the hipSOLVER functions with the cuSOLVER backend. We can make hipSOLVER robust to nullptrs even if cuSOLVER is not.

jzuniga-amd:
I am not sure about modifying the behaviour of the CUDA backend; if they don't care about null input pointers and return success (I guess is a quick return), then hipSOLVER should return success as well. 
Maybe the solution is to skip the null-pointer test for the CUDA backend (or check that the returned status is success), but keep testing for the HIP backend. Would that be possible?

cgmb:
Given that passing null pointers causes future tests to fail, it sounds like cuSOLVER is simply not doing any checks at all. If you want to only do them for the rocm backend, you could put the tests in an `#ifdef __HIP_PLATFORM_AMD__` section.

tfalders:
`__HIP_PLATFORM_AMD__` did not appear to be defined when I tried to run the tests, so I used `__HIP_PLATFORM_NVCC__` instead. The cmake files indicate it will always be defined when we use the cuSOLVER backend.

cgmb:
`__HIP_PLATFORM_HCC__` is being replaced with `__HIP_PLATFORM_AMD__` and `__HIP_PLATFORM_NVCC__` is being replaced with `__HIP_PLATFORM_NVIDIA__`.

We could probably use `#if defined(__HIP_PLATFORM_HCC__) || defined(__HIP_PLATFORM_AMD__)` for compatibility.

jzuniga-amd (approved):
LGTM

PR #13, opened 2021-04-13
Leave AMDGPU_TARGETS unset

cgmb:
hipSOLVER does not need to set AMDGPU_TARGETS. It has no kernels
itself, so this variable does not control anything within the library.


tfalders (approved):

jzuniga-amd (approved):
LGTM

PR #14, opened 2021-04-16
Added potrf_batched

tfalders:
Adds potrf_batched to hipSOLVER


clients/include/testing_potrf.hpp
```diff
@@ -133,14 +133,14 @@ void potrf_initData(const hipsolverHandle_t   handle,
     }
 }
```
jzuniga-amd, line 134: Here you also need to include the fix on the bad_arg test for the nullptrs on the cuda side...

jzuniga-amd (approved):
LGTM
(Just make sure you are applying all the fixes and improvements from the other PRs to all functions)

PR #15, opened 2021-05-11
Enable ccache on CI by default

cgmb:
This PR makes use of ccache the default setting for CI builds. It can be disabled for PRs using the label `ci:no-ccache`.

This doesn't provide much value for hipSOLVER, as its builds are quite fast anyway. I mostly just want to check that this CI functionality is working properly for the builds that support both NVIDIA and AMD.


jzuniga-amd (approved):
LGTM

tfalders (approved):
LGTM!

PR #16, opened 2021-05-12
Updated documentation with special considerations for rocSOLVER

tfalders:


jzuniga-amd (approved):
LGTM

jzuniga-amd:
BTW, I was thinking that now that this feature (the library managing workspace when the user passes `work = nullptr`) is official and documented, we should be testing it as well ??

Maybe some mechanism similar to what we have for bad_args, ie. for some particular test case, run the test normally once and then repeat it passing `work=nullptr` ??

rocsolver managing workspace is being tested with the other functions that do not have `work` as input argument, but it might be a good idea to do the test explicitly for those functions requiring workspace from the user, what do you think?

tfalders:
> BTW, I was thinking that now that this feature (the library managing workspace when the user passes `work = nullptr`) is official and documented, we should be testing it as well ??

Yeah, that sounds like a good idea; I'll take a look at adding it in. I might also update the documentation more to note the difference between how the info argument is handled by cuSOLVER and rocSOLVER.

tfalders:
I will probably add tests for null `work` pointers in a separate PR and leave this one specifically for documentation changes. Re-requesting review as I have expanded the documented behaviour.

README.md
```diff
@@ -87,7 +87,16 @@ hipsolverSgetrf(hipsolverHandle_t handle,
...
 ```
 
-## Supported Fuctionality
+## Special Considerations with the rocSOLVER Backend
+Due to differences in implementation and API design between rocSOLVER and cuSOLVER, there are special considerations that should be taken into account when using hipSOLVER with the rocSOLVER backend.
+
+While many hipSOLVER functions (modeled after cuSOLVER functions) take a workspace pointer and size as arguments, rocSOLVER maintains its own internal device workspace by default. In order to take advantage of this feature, users may pass a null pointer for the `work` argument of any function when using the rocSOLVER backend, and the workspace will be automatically managed behind-the-scenes.
+
+Note that several functions - namely gesvd, getrs, and potrfBatched - will always use rocSOLVER's internal device workspace management. This may cause performance issues if combined with function calls that receive non-null `work` pointers, as the internal workspace will flip-flop between the user-provided and automatically allocated workspaces. It is recommended for programs that call gesvd, getrs, or potrfBatched to pass null pointers to the `work` arguments of all other functions.
```
cgmb, line 95: A single hypen is too small for an em dash. You can use Unicode:

> Note that several functions—namely gesvd, getrs, and potrfBatched—will always use rocSOLVER's

Or a few hyphens, if you want to stick to ASCII:
> Note that several functions---namely gesvd, getrs, and potrfBatched---will always use rocSOLVER's

FWIW, according to the Chicago and Harvard style manuals, there should be no spaces around the em dash. Adding spaces is more common in Canadian / UK style, but the ROCm docs follow US English style guides. (That said, even US authors often do this differently, so it's not that big of a deal.)

README.md
```diff
@@ -87,7 +87,16 @@ hipsolverSgetrf(hipsolverHandle_t handle,
                 int*              devInfo);
 ```
 
-## Supported Fuctionality
+## Special Considerations with the rocSOLVER Backend
+Due to differences in implementation and API design between rocSOLVER and cuSOLVER, there are special considerations that should be taken into account when using hipSOLVER with the rocSOLVER backend.
```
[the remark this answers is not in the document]
tfalders: I disagree. The rocSOLVER backend is, in a sense, jury-rigged to fit into the cuSOLVER API design, and this causes issues that the cuSOLVER backend does not have. I don't think it makes sense to frame the ignored info arguments or flip-flopping between memory models as differences between the backends - they are unfortunate side effects caused by mapping one library design into another.

I'll see if I can play around with the wording a little bit, but I think it would be best to keep the same framing.

README.md
jzuniga-amd, line 95: We may add more functions that do not receive `work` pointer. So, I will re-phrase these paragraph to something like this:

"Note that some functions that do not get a `work` array as an argument, will always use rocSOLVER's internal device workspace management (gesvd, getrs, and potrfBatched, for instance). This may cause performance issues if combined with function calls that receive non-null `work` pointers, as the internal workspace will flip-flop between the user-provided and automatically allocated workspaces. In these cases, it is recommended to always pass null pointers to the `work` arguments for the functions that need it."

jzuniga-amd (approved):
LGTM
Just a minor change...

cgmb (approved):

PR #17, opened 2021-05-12
Added ormqr

tfalders:
Adds ormqr to hipSOLVER


jzuniga-amd (approved):
LGTM

clients/common/hipsolver_datatype2string.cpp
```diff
@@ -88,3 +100,19 @@ hipsolverFillMode_t char2hipsolver_fill(char value)
...
+    case 'L':
+        return HIPSOLVER_SIDE_LEFT;
+    case 'R':
+        return HIPSOLVER_SIDE_RIGHT;
+    case 'l':
+        return HIPSOLVER_SIDE_LEFT;
+    case 'r':
+        return HIPSOLVER_SIDE_RIGHT;
+    }
+    return HIPSOLVER_SIDE_LEFT;
```
cgmb, line 117: If there are two identical cases, I would use fallthrough. e.g.
```
    case 'l':
    case 'L':
        return HIPSOLVER_SIDE_LEFT;
    case 'r':
    case 'R':
        return HIPSOLVER_SIDE_RIGHT;
    }
```
The final `return HIPSOLVER_SIDE_LEFT;` concerns me a little. It appears to be a case that should never happen. It would be nice to have some warning if it ever does, either via an assertion or an exception.

cgmb (approved):
Just a minor comment. Address at your leisure.

PR #18, opened 2021-05-12
Added gesvd

tfalders:
Adds gesvd to hipSOLVER

### DIFF
--- a/clients/gtest/gesvd_gtest.cpp
+++ b/clients/gtest/gesvd_gtest.cpp
@@ -26,6 +26,9 @@ typedef std::tuple<vector<int>, vector<int>> gesvd_tuple;
 // if leftsv (rightsv) = 2 then compute all orthogonal matrix
 // if leftsv (rightsv) = 3 then no singular vectors are computed
 
+// case when m = -1, n = 1, and rightsv = leftsv = 3 will also execute the bad
+// arguments test (null handle, null pointers and invalid values)
+
 // for checkin_lapack tests
 const vector<vector<int>> size_range = {
     // invalid
@@ -137,6 +140,10 @@ protected:
     void run_tests()
     {
         Arguments arg = gesvd_setup_arguments(GetParam());
+
+        if(arg.peek<rocblas_int>("m") == -1 && arg.peek<rocblas_int>("n") == 1
+           && arg.peek<char>("jobu") == 'N' && arg.peek<char>("jobv") == 'N')
+            testing_gesvd_bad_arg<FORTRAN, BATCHED, STRIDED, T>();
 
         arg.batch_count = 1;
         testing_gesvd<FORTRAN, BATCHED, STRIDED, T>(arg);
--- a/clients/gtest/gesvd_gtest.cpp
+++ b/clients/gtest/gesvd_gtest.cpp
@@ -61,16 +61,16 @@ const vector<vector<int>> opt_range = {
     {0, 0, 0, 0, 2},
     {0, 0, 0, 0, 1}};
 
-// for daily_lapack tests
-const vector<vector<int>> large_size_range = {{120, 100, 0}, {300, 120, 0}, {300, 120, 1}};
+// // for daily_lapack tests
+// const vector<vector<int>> large_size_range = {{120, 100, 0}, {300, 120, 0}, {300, 120, 1}};
 
-const vector<vector<int>> large_opt_range = {{0, 0, 0, 3, 3},
-                                             {1, 0, 0, 0, 1},
-                                             {0, 1, 0, 1, 0},
-                                             {0, 0, 1, 1, 1},
-                                             {0, 0, 0, 3, 0},
-                                             {0, 0, 0, 1, 3},
-                                             {0, 0, 0, 3, 2}};
+// const vector<vector<int>> large_opt_range = {{0, 0, 0, 3, 3},
+//                                              {1, 0, 0, 0, 1},
+//                                              {0, 1, 0, 1, 0},
+//                                              {0, 0, 1, 1, 1},
+//                                              {0, 0, 0, 3, 0},
+//                                              {0, 0, 0, 1, 3},
+//                                              {0, 0, 0, 3, 2}};
 
 Arguments gesvd_setup_arguments(gesvd_tuple tup)
 {
@@ -193,15 +193,15 @@ TEST_P(GESVD_FORTRAN, __double_complex)
     run_tests<false, false, rocblas_double_complex>();
 }
 
-INSTANTIATE_TEST_SUITE_P(daily_lapack,
-                         GESVD,
-                         Combine(ValuesIn(large_size_range), ValuesIn(large_opt_range)));
+// INSTANTIATE_TEST_SUITE_P(daily_lapack,
+//                          GESVD,
+//                          Combine(ValuesIn(large_size_range), ValuesIn(large_opt_range)));
 
 INSTANTIATE_TEST_SUITE_P(checkin_lapack, GESVD, Combine(ValuesIn(size_range), ValuesIn(opt_range)));
 
-INSTANTIATE_TEST_SUITE_P(daily_lapack,
-                         GESVD_FORTRAN,
-                         Combine(ValuesIn(large_size_range), ValuesIn(large_opt_range)));
+// INSTANTIATE_TEST_SUITE_P(daily_lapack,
+//                          GESVD_FORTRAN,
+//                          Combine(ValuesIn(large_size_range), ValuesIn(large_opt_range)));
 
 INSTANTIATE_TEST_SUITE_P(checkin_lapack,
                          GESVD_FORTRAN,
--- a/clients/gtest/gesvd_gtest.cpp
+++ b/clients/gtest/gesvd_gtest.cpp
@@ -1,0 +1,208 @@
+/* ************************************************************************
+ * Copyright 2020-2021 Advanced Micro Devices, Inc.
+ *
+ * ************************************************************************ */
+
+#include "testing_gesvd.hpp"
+
+using ::testing::Combine;
+using ::testing::TestWithParam;
+using ::testing::Values;
+using ::testing::ValuesIn;
+using namespace std;
+
+typedef std::tuple<vector<int>, vector<int>> gesvd_tuple;
+
+// each size_range vector is a {m, n, fa};
+// if fa = 0 then no fast algorithm is allowed
+// if fa = 1 fast algorithm is used when possible
+
+// each opt_range vector is a {lda, ldu, ldv, leftsv, rightsv};
+// if ldx = -1 then ldx < limit (invalid size)
+// if ldx = 0 then ldx = limit
+// if ldx = 1 then ldx > limit
+// if leftsv (rightsv) = 0 then overwrite singular vectors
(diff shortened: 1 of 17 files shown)


clients/rocsolvercommon/rocsolver_test.hpp
```diff
@@ -53,17 +53,17 @@ inline void rocsolver_bench_output(T arg, Ts... args)
...
-// template <typename T, std::enable_if_t<is_complex<T>, int> = 0>
-// inline T sconj(T scalar)
-// {
-//     return std::conj(scalar);
-// }
+template <typename T, std::enable_if_t<is_complex<T>, int> = 0>
+inline T sconj(T scalar)
+{
+    return std::conj(scalar);
+}
```
cgmb, line 66: IIRC, Juan originally added sconj as a workaround for some compiler or library issue in ROCm 3.9. If that issue has been resolved, perhaps we can just use `std::conj`?

clients/include/hipsolver_fortran.f90
```diff
@@ -153,6 +153,167 @@ function hipsolverZgeqrfFortran(handle, m, n, A, lda, tau, work, lwork, info) &
...
+        use iso_c_binding
+        use hipsolver_enums
+        implicit none
+        type(c_ptr), value :: handle
+        integer(c_int), value :: m
+        integer(c_int), value :: n
+        type(c_ptr), value :: lwork
+        integer(c_int) :: res
+        res = hipsolverSgesvd_bufferSize(handle, m, n, lwork)
+    end function hipsolverSgesvd_bufferSizeFortran
```
cgmb, line 169: This is more a generic question about hipsolver than one specific to SVD, but does anything use the Fortran API? If not, we might want to add a Fortran sample.

tfalders: By "anything", do you mean client code? The client code tests it in an extremely roundabout way. The tests will call functions in hipsolver.hpp with a FORTRAN bool, which will call functions in hipsolver_fortran.hpp with a Fortran suffix, which extern to Fortran code in hipsolver_fortran.f90, which itself will call the library Fortran code in hipsolver_module.f90.

I guess the intent is specifically to make sure we can call our Fortran code from other Fortran code - everything else is just the C++ infrastructure to get us to that point. A sample would have to be written in Fortran as well, which I wouldn't know how to do.

clients/rocsolvercommon/rocsolver_test.hpp
tfalders: Switching it over seems to work fine. We'll probably want to make the same change to rocSOLVER.

clients/gtest/gesvd_gtest.cpp
```diff
@@ -0,0 +1,215 @@
...
+const vector<vector<int>> size_range = {
+    // invalid
+    {-1, 1, 0},
+    {1, -1, 0},
+    // normal (valid) samples
+    {1, 1, 0},
+    {20, 20, 0},
+    {40, 30, 0},
+    {60, 30, 0},
+    {60, 30, 1}};
```
jzuniga-amd, line 42: You always call rocsolver_gesvd in-place, so this last test case is repeated; it could be removed.
(Not sure yet if it is a good idea to always call in place, though... see my other related comments)

library/src/hcc_detail/hipsolver.cpp
```diff
@@ -697,6 +714,219 @@ catch(...)
     return exception2hip_status();
 }
 
+/******************** GESVD ********************/
+hipsolverStatus_t hipsolverSgesvd_bufferSize(hipsolverHandle_t handle, int m, int n, int* lwork)
+try
+{
+    *lwork = 0;
+    return HIPSOLVER_STATUS_NOT_SUPPORTED;
```
jzuniga-amd, line 722: Why is this not supported?
Users should be able to pass a pointer to work of size returned by this function (ie. rocsolver_gesvd_getMemorySize) with both backends

library/src/hcc_detail/hipsolver.cpp
```diff
@@ -697,6 +714,219 @@ catch(...)
...
+                                               char2rocblas_svect(jobv),
+                                               m,
+                                               n,
+                                               A,
+                                               lda,
+                                               S,
+                                               U,
+                                               ldu,
+                                               V,
+                                               ldv,
```
jzuniga-amd, line 794: You should set the workspace to user-owned (rocblas_set_workspace())...
Default to auto managing only if work==nullptr as with the other functions accepting workspace provided by user.

library/src/hcc_detail/hipsolver.cpp
```diff
@@ -697,6 +714,219 @@ catch(...)
...
+                                               rwork,
+                                               rocblas_inplace,
```
jzuniga-amd, line 796: Using in-place should be, in most cases, slower than out-of-place...
So, there are more chances for this function on the HIP side to be always slower than on CUDA side. I would say always use out-of-place instead, but we don't want to require too much more memory on HIP than on CUDA. So I suggest you check, for some given sizes, the memory required by cuSOLVER and rocSOLVER in both cases, in-place and out-of-place. If out-of-place is compared to what cuSOLVER needs, then switch to always compute out-of-place.

jzuniga-amd (changes requested):

clients/rocsolvercommon/rocsolver_test.hpp
jzuniga-amd: It works because we are injecting `conj` for real cases in the std namespace here `clients/include/complex.hpp`
rocBLAS does not do that for rocblas_complex types, so this won't work in rocSOLVER.
For some reason, std::conj does not like rocblas_complex types...

library/src/hcc_detail/hipsolver.cpp
tfalders: I left this as unsupported for now because the only two arguments to gesvd's bufferSize function are `m` and `n`. I wasn't sure how to get the maximum needed memory size without querying rocsolver_gesvd using every possible combination of `left_svect` and `right_svect`.

library/src/hcc_detail/hipsolver.cpp
tfalders: I'm planning to do that if/when we support the bufferSize function for gesvd. Otherwise, there's a risk that the user could provide a workspace that is not sufficiently large for gesvd and cause a memory fault.

library/src/hcc_detail/hipsolver.cpp
tfalders: I seem to recall having a reason for making it in-place, but I've forgotten what it was. I'll try switching it over to out-of-place and comparing memory sizes, and perhaps I'll remember what my reasoning was.

library/src/hcc_detail/hipsolver.cpp
jzuniga-amd: oh, I see!

My suggestion is to add `left_svect` and `right_svect` to the bufferSize APIs, and ignore them on the CUDA side. The call won't be exactly as in cuSOLVER, but it will work (better to have extra parameters than missing parameters).
I think this is better than having the same user code falling on the HIP side because the bufferSize API will return not_supported.

Alternatively, returning lwork=0 and a success status should be fine... 
If lwork is zero, and you do `rocblas_set_workspace` on the implementation of the gesvd function, the memory model will be switched to automatic anyways.

What do you think?

tfalders: I would rather not deviate from the cuSOLVER API by adding new arguments, as my understanding is that the point of hipBLAS and hipSOLVER is that the user should just be able to hipify their cuda code and have it run on our hardware without issue. The easier it is to turn cuSOLVER code into hipSOLVER code, the better.

I could change the returned status to be a success, sure. My only concern is that will sort of hide the fact that we always use the internal workspace with gesvd and never the user-provided workspace. On that note, I'd still prefer not to use `rocblas_set_workspace` in gesvd. If a user creates a workspace of size `max(getrf_bufferSize, gesvd_bufferSize)`, for example, it might not be large enough since `gesvd_bufferSize` is not giving the correct size. If the user passes that workspace to gesvd with `lwork != 0`, and we call `rocblas_set_workspace` on it, that will cause problems.

jzuniga-amd: No, the main purpose of hipBLAS and hipSOLVER is not hipify code (there are other tools like hipify-clang for that). hipBLAS/SOLVER are marshalling libraries; they should provide an interface that does not require the user code to change regardless of what backend (cuda or hip) is used. Ideally, to facilitate the transition of cuda libraries' users to rocm libraries, we want this interface to be as similar to cuBLAS/SOLVER as possible, but it is not a requirement.  

The requirement is that the same code, let's say:

int lwork;
T *work;
hipsolverSgesvd_bufferSize(...,*lwork);
hipMalloc(..., work, work);
hipsolverSgesvd(..., work, lwork);

works OK with both backends. This won't be the case if the user checks returned status of the buffeSize API, for example; or, as you mentioned if we return lwork=0 but the user then does lwork=max(...) with the buffer size of another function, etc (there are many ways this could fail)...

IMO, the safest bet is to have the bufferSize API receiving the svects as parameters and letting both backends do their thing.

tfalders: Interesting. If we're going to be allowing changes from the cuSOLVER API, there are other changes I would like to make, such as adding workspace arguments to getrs and potrfBatched, as well as adding an lwork argument to getrf. I'll have to update the documentation about the special considerations.

clients/rocsolvercommon/rocsolver_test.hpp
cgmb: Oh, I see what's going on now. The actual [`std::conj`](https://en.cppreference.com/w/cpp/numeric/complex/conj) has a specialization that takes a real input (by value), but it returns `std::complex`. We have defined an overload of `std::complex` that takes a real input (by reference) and returns a real output. That's very confusing.

The [section of clients/include/complex.hpp which does that](https://github.com/ROCmSoftwarePlatform/hipSOLVER/blob/develop/clients/include/complex.hpp#L156-L207) is not good. There's very few cases in which it's valid to insert something into the `std::` namespace, and I would not be surprised if that code is causing half the problems that `sconj` is solving.

My suggestion: when we swap out `hipsolverComplex` for `hipFloatComplex`, we should drop all those overloads from the `std` namespace. We can define our own functions on an as-needed basis.

library/src/hcc_detail/hipsolver.cpp
jzuniga-amd: Adding lwork to getrf is not a bad idea!... This way it would be cleaner on the hip side; no need to do the memory query on the function implementation. On the cuda side, lwork could be ignored.

On the other hand, I don't see why adding workspace to getrs or potrf calls, neither rocSOLVER nor cuSOLVER really need it.  rocSOLVER will manage workspace automatically, and cuSOLVER will do whatever it does.

tfalders: I want to add workspaces to getrs and potrfBatched so that we can avoid switching between user-provided and internal workspaces when the user wants to provide the workspace. As noted in the documentation, this switching can impact performance.

jzuniga-amd (approved):
LGTM

cgmb (approved):

PR #21, opened 2021-05-17
Added orgqr

tfalders:
Adds orgqr to hipSOLVER

### DIFF
--- a/library/src/hcc_detail/hipsolver.cpp
+++ b/library/src/hcc_detail/hipsolver.cpp
@@ -173,6 +173,227 @@ catch(...)
     return exception2hip_status();
 }
 
+/******************** ORGQR/UNGQR ********************/
+hipsolverStatus_t hipsolverSorgqr_bufferSize(
+    hipsolverHandle_t handle, int m, int n, int k, float* A, int lda, float* tau, int* lwork)
+try
+{
+    size_t sz;
+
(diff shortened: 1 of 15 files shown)


jzuniga-amd (approved):
LGTM

library/src/hcc_detail/hipsolver.cpp
```diff
@@ -173,6 +173,227 @@ catch(...)
...
+                                  int*              devInfo)
+try
+{
+    if(work != nullptr)
+        rocblas_set_workspace((rocblas_handle)handle, work, lwork);
+    else
+    {
+        if(!rocblas_is_managing_device_memory((rocblas_handle)handle))
+            rocblas_set_workspace((rocblas_handle)handle, nullptr, 0);
+    }
```
cgmb, line 295: It's rather trivial, but would `else if` make sense here?

cgmb (approved):
Minor comment / question. Does not need to be addressed.

library/src/hcc_detail/hipsolver.cpp
tfalders: That would look nicer, I think. It'll affect all the other functions too, so I'll probably wait until they're all in and open a new PR. Should be able to do a search-and-replace.

PR #25, opened 2021-05-26
API changes

tfalders:
As discussed in https://github.com/ROCmSoftwarePlatform/hipSOLVER/pull/18#discussion_r639136057, I am making the following API changes to hipSOLVER:

1. getrf now takes an `lwork` argument
2. getrs now takes `work` and `lwork` arguments (this required the addition of a getrs_bufferSize method)
3. potrfBatched now takes `work` and `lwork` arguments (this required the addition of a potrfBatched_bufferSize method)

As with the removal of HIPSOLVER_FILL_MODE_FULL in https://github.com/ROCmSoftwarePlatform/hipSOLVER/pull/4, I think that we are early enough in the development process for hipSOLVER that a deprecation process for getrf is currently unnecessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,18 +17,21 @@
     - hipsolverSgeqrf_bufferSize, hipsolverDgeqrf_bufferSize, hipsolverCgeqrf_bufferSize, hipsolverZgeqrf_bufferSize
     - hipsolverSgeqrf, hipsolverDgeqrf, hipsolverCgeqrf, hipsolverZgeqrf
   - getrs
+    - hipsolverSgetrs_bufferSize, hipsolverDgetrs_bufferSize, hipsolverCgetrs_bufferSize, hipsolverZgetrs_bufferSize
     - hipsolverSgetrs, hipsolverDgetrs, hipsolverCgetrs, hipsolverZgetrs
   - potrf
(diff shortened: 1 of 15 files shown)


jzuniga-amd (commented):
Looks good...
I just made some minor suggestions for the documentation in a PR to your branch. Please have a look.

jzuniga-amd (approved):
LGTM

cgmb (commented):
If I understand the conversation that spawned this change correctly, the hipSOLVER API may be slightly different from either cuSOLVER or rocSOLVER. In that case, should we be creating individual function docs? It might be useful even if only to list the parameters and refer to the associated ROCm / CUDA functions for more details. (Or maybe it's still unnecessary?)

Even if we should add those docs, there's no need to do that immediately. I'm just wondering if that's worth creating eventually.

tfalders:
> If I understand the conversation that spawned this change correctly, the hipSOLVER API may be slightly different from either cuSOLVER or rocSOLVER.

Correct.

> In that case, should we be creating individual function docs? It might be useful even if only to list the parameters and refer to the associated ROCm / CUDA functions for more details. (Or maybe it's still unnecessary?)

Yeah, @doctorcolinsmith once suggested to add the API to the hipSOLVER docs, even if it only shows the method names and signatures. I haven't gotten around to it yet, but it's something I would like to add.

cgmb (approved):
👍

PR #26, opened 2021-05-26
Added syevd & sygvd

tfalders:
Adds syevd and sygvd to hipSOLVER.


cgmb (commented):
Why is there a version bump in this PR?

tfalders:
> Why is there a version bump in this PR?

syevd and sygvd don't exist in rocSOLVER 3.13.

cgmb (approved):
LGTM

library/src/hcc_detail/hipsolver.cpp
```diff
@@ -2902,6 +2958,664 @@ catch(...)
...
+                                  int                 n,
+                                  float*              A,
+                                  int                 lda,
+                                  float*              W,
+                                  float*              work,
+                                  int                 lwork,
+                                  int*                devInfo)
+try
+{
+    float* E = work;
```
jzuniga-amd, line 3126: how is this going to work if a user passes a nullptr because wanted to use rocsolver automatic workspace management?
You cannot pass E as nullptr to rocsolver_syevd

library/src/hcc_detail/hipsolver.cpp
```diff
@@ -2902,6 +2958,664 @@ catch(...)
...
+                                             n,
+                                             nullptr,
+                                             lda,
+                                             nullptr,
+                                             nullptr,
+                                             nullptr);
+    rocblas_stop_device_memory_size_query((rocblas_handle)handle, &sz);
+
+    // space for E array
+    sz += sizeof(float) * n;
```
jzuniga-amd, line 2987: And what would happen if a user never calls this function because it wants to use rocsolver automatic reallocation?
How would you allocate space for E?

library/src/hcc_detail/hipsolver.cpp
```diff
@@ -2902,6 +2958,664 @@ catch(...)
...
+                                               W,
+                                               E,
+                                               devInfo));
+}
+catch(...)
+{
+    return exception2hip_status();
+}
+
+/******************** SYTRD/HETRD ********************/
```
jzuniga-amd, line 3266: wrong comment!!

jzuniga-amd (commented):
Looks good, but there are some things we should clarify...

library/src/hcc_detail/hipsolver.cpp
tfalders: Oh, you're right! That complicates things...

I guess we have a couple options.
1. Add E as an argument to the function. On the CUDA side it will do nothing.
2. If work is a null pointer, use `rocblas_device_malloc` to allocate internal space for E. I'm not sure how `rocblas_device_malloc` will respond to being used in both hipSOLVER and rocSOLVER.
3. Create a private API for syevd and sygvd in rocSOLVER that does not need E.

library/src/hcc_detail/hipsolver.cpp
tfalders: Oops! Good catch, thanks.

clients/include/testing_syevd_heevd.hpp
```diff
@@ -0,0 +1,779 @@
...
+    using S = decltype(std::real(T{}));
+
+    // get arguments
+    hipsolver_local_handle handle;
+    char                   evectC = argus.get<char>("jobz");
+    char                   uploC  = argus.get<char>("uplo");
+    int                    n      = argus.get<int>("n");
+    int                    lda    = argus.get<int>("lda", n);
+    int                    stA    = argus.get<int>("strideA", lda * n);
+    int                    stD    = argus.get<int>("strideW", n);
```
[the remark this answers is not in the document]
tfalders: In the API, I adopted cuSOLVER's naming of `W` instead of `D`. I figured that if we were to add a strided batch version, that the argument would come into hipsolver-bench as `strideW` instead of `strideD`.

I suppose I could just change it to use `D` as the name instead of `W`. It would cause less confusion for us.

library/src/hcc_detail/hipsolver.cpp
jzuniga-amd: I would try option 2 first. I believe it should work... at the beginning, before rocblas team added the `rocblas_device_malloc` wrapper, we were allocating workspace arrays one by one. 
If this does not work, I suggest option 1 then. 

(just to clarify, in rocSOLVER we cannot have E as part of the workspace because when info!=0, this array will contain information of the equivalent resulting system that could be useful in some applications. If E is part of the internally managed workspace we cannot guarantee it will persist after the call to the rocsolver function returns)

tfalders: Option 2 generally seems to work well, but I was able to trigger a failure case by allocating a large amount of memory with `rocblas_device_malloc` after E had already been allocated (the pointer to E is likely invalidated by rocBLAS reallocating the memory space). I'm not sure how big the problem has to be in order to trigger this failure case, since daily_lapack seemed to run fine.

I'd prefer to avoid using option 1. Perhaps option 2 can be salvaged by using `hipMalloc` instead of `rocblas_device_malloc`. I know the philosophy of hipBLAS does not allow calls to `hipMalloc`; do we want to follow the same philosophy in hipSOLVER?

tfalders:
After discussing with @jzuniga-amd, we decided to manually specify the workspace size if a null workspace is provided to hipSOLVER functions. This should allow us to allocate E using `rocblas_device_malloc` without causing a failure, and also addresses the possibility of rocBLAS disabling automatic reallocation. Note that CI won't pass until https://github.com/ROCmSoftwarePlatform/rocBLAS-internal/pull/716 is merged.

I'll re-request a review from @cgmb, since this is a large change. Only hcc_detail/hipsolver.cpp has really been affected, so you can limit your review to that file.

library/src/hcc_detail/hipsolver.cpp
```diff
@@ -209,6 +211,27 @@ hipsolverStatus_t rocblas2hip_status(rocblas_status_ error)
...
+    } while(0)
+
+inline rocblas_status hipsolverManageWorkspace(rocblas_handle handle, int lwork)
+{
+    size_t current_size;
+    if(rocblas_is_user_managing_device_memory(handle))
+        rocblas_get_device_memory_size(handle, &current_size);
+
+    if(lwork > current_size)
+        return rocblas_set_device_memory_size(handle, lwork);
```
cgmb, line 229: `lwork` is user-supplied, and is implicitly converted from `int` to `size_t` for this call. If the user has supplied a negative value, the result will be an allocation with a nonsense size (probably very large). That might be fine. It's garbage-in-garbage-out, after all. Just checking whether that's the intended design of the API.

Even if there's no validity check being done, I would recommend making the cast to `size_t` explicit. It will make the conversion more obvious to future developers, and you can avoid a signed / unsigned comparison by putting it before the `lwork > current_size` comparison and using the converted value. The C++ integer promotion rules are surprisingly complicated, so it's nice to avoid the complexity by matching types exactly and making conversions explicit.

library/src/hcc_detail/hipsolver.cpp
cgmb: > I was able to trigger a failure case by allocating a large amount of memory with `rocblas_device_malloc` after E had already been allocated (the pointer to E is likely invalidated by rocBLAS reallocating the memory space)

That's unfortunate. I don't think there's any reason why that would have to be the case, but the rocblas allocator is probably too complex to change in time the next release cut-off.

jzuniga-amd (approved):
LGTM...
Nice, good job, Troy.

cgmb (approved):
:+1:

tfalders:
Some syevd/heevd and sygvd/hegvd tests were failing because of incorrect status code returns. These should be fixed now:
1. syevd/heevd and sygvd/hegvd were not checking the returned status of bufferSize functions, which was causing some errors to be reported as failed allocs from hipsolverManageWorkspace.
2. bufferSize functions were allowed to overwrite error codes from the queried function with HIPSOLVER_STATUS_INTERNAL_ERROR. Error codes should no longer be overridden.

cgmb:
I liked the cleanup in 7d0f318071d566659211e5e51ed23f92bee348ff. It looked like you might have just missed changing `GESVD`. In any case, nice catch on the error handling.

tfalders:
> I liked the cleanup in [7d0f318](https://github.com/ROCmSoftwarePlatform/hipSOLVER/commit/7d0f318071d566659211e5e51ed23f92bee348ff). It looked like you might have just missed changing `GESVD`. In any case, nice catch on the error handling.

Thanks! I'm planning to reopen the cleanup but as a separate PR. I wasn't comfortable making a whole bunch of changes + fixing those changes without a review.

PR #29, opened 2021-06-08
bufferSize clean up

tfalders:
Some basic changes to the bufferSize functions, plus a nullptr check for lwork.

### DIFF
--- a/library/src/hcc_detail/hipsolver.cpp
+++ b/library/src/hcc_detail/hipsolver.cpp
@@ -309,17 +309,17 @@ try
     size_t sz;
 
     rocblas_start_device_memory_size_query((rocblas_handle)handle);
-    rocblas_status status = rocsolver_sorgbr(
-        (rocblas_handle)handle, hip2rocblas_side2storev(side), m, n, k, nullptr, lda, nullptr);
+    hipsolverStatus_t status = rocblas2hip_status(rocsolver_sorgbr(
+        (rocblas_handle)handle, hip2rocblas_side2storev(side), m, n, k, nullptr, lda, nullptr));
     rocblas_stop_device_memory_size_query((rocblas_handle)handle, &sz);
 
-    bool valid
-        = (status == rocblas_status_size_unchanged || status == rocblas_status_size_increased);
-    if(valid && sz > INT_MAX)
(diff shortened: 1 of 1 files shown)


jzuniga-amd (approved):
LGTM

library/src/hcc_detail/hipsolver.cpp
```diff
@@ -306,20 +306,24 @@ hipsolverStatus_t hipsolverSorgbr_bufferSize(hipsolverHandle_t   handle,
                                              int*                lwork)
 try
 {
+    if(lwork == nullptr)
+        return HIPSOLVER_STATUS_INVALID_VALUE;
+
+    *lwork = 0;
```
cgmb, line 312: I'm curious why we're setting this to zero when there's an error. Is that to match existing CUDA or LAPACK behaviour?

cgmb (commented):
It looks great. I just have a question about setting lwork in the error case.

library/src/hcc_detail/hipsolver.cpp
tfalders: This is to address the test failure I was getting from gesvd. Since we don't initialize `size_W` or `lwork` or check for bufferSize errors in the tests, `lwork` was taking a random (large) value and we were getting an allocation errors. Alternatively, I could have changed something in the tests, but I thought this would increase the safety of our code.

cgmb: If CUDA doesn't guarantee this, won't it be a problem in when we run the tests on the CUDA build?

tfalders: Oh, good point! I'll check what CUDA does. If they don't set lwork in the case of an error, I'll update the tests so that `size_W` is initialized to zero.

tfalders: It looks like CUDA does set `lwork` even if the problem size is invalid (for gesvd, at least). Interestingly, it's a non-zero value.

cgmb: That is rather odd. I feel like it would be better if the tests checked for errors, but we can probably merge as-is.

cgmb (approved):

PR #30, opened 2021-06-10
Add address sanitizer build option

eidenyoshida:


eidenyoshida:
@cgmb Wondering if I could get your help figuring out this issue. When building with the address sanitizer flag, I get the following issue

[  8%] Linking Fortran shared library libhipsolver_fortran.so
collect2: fatal error: cannot find 'ld'
compilation terminated.
library/src/CMakeFiles/hipsolver_fortran.dir/build.make:86: recipe for target 'library/src/libhipsolver_fortran.so' failed

Suspect it has something to do with gfortran. I'm getting the same issue on hipBLAS address sanitizer. Not quite sure how to resolve it

cgmb:
If you build with the `VERBOSE` environment variable set (e.g. `make VERBOSE=1 -C build/release`), you can see the actual commands used and it's a lot easier to track down problems. You can copy/paste/modify the build commands for individual files, and try different variations until you find one that works.

In this case, the problem seems to be `-fuse-ld=lld`. `gfortran` is used for the linking of the Fortran shared library, and it doesn't like that flag. We could probably try a different approach that only adds these flags to selected targets, so they're not used for Fortran code.

cgmb:
I'll send you a PR with a most robust fix for that problem. It's possible to conditionally add flags based on the compile or link language.

cgmb:
The debug build actually needs the same change as rocSOLVER with respect to `_FORTIFY_SOURCE`: https://github.com/ROCmSoftwarePlatform/rocSOLVER/blob/7f886a16394dfa0cf1b34557721cbed49b4a6717/cmake/armor-config.cmake#L21

jzuniga-amd (approved):
LGTM

PR #31, opened 2021-06-10
Use hip complex types

tfalders:
Replaces `hipsolverComplex` with `hipFloatComplex` and `hipsolverDoubleComplex` with `hipDoubleComplex` in the library. I was not able to replace these in the client code, so `hipsolverComplex` and `hipsolverDoubleComplex` have been moved to `complex.hpp`. `hipsolver.hpp` now casts pointers of the original complex types to pointers of the hip complex types.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -53,6 +53,7 @@
 
 ### Removed
 - Removed unused HIPSOLVER_FILL_MODE_FULL enum value.
+- Removed hipsolverComplex and hipsolverDoubleComplex from the library. Use hipFloatComplex and hipDoubleComplex instead.
 
 
 ## [hipSOLVER 0.1.0 for ROCm 4.2]
--- a/clients/include/complex.hpp
+++ b/clients/include/complex.hpp
@@ -7,6 +7,78 @@
 #include "hipsolver.h"
 #include <complex>
 
+typedef struct hipsolverComplex
+{
+private:
+    float x, y;
+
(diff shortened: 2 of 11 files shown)


cgmb:
`HIPSOLVER_CLANG_STATIC` appears unused and undocumented. Should we remove that, too?

tfalders:
> edit: Oh, it's used in the clients. Not clear why it's in the public API, though.

This is another one of those things I copied from hipBLAS. The comment indicates it's a workaround for some bug. I'm not sure if it's been fixed yet.

cgmb:
It was introduced here in hipBLAS: https://github.com/ROCmSoftwarePlatform/hipBLAS/commit/c09786ace8bc68df94750594929c4e6322001d63.

The bug is in a C++14 feature used with the custom complex type (template variables), so we can move that inside the library along with the custom complex type.

jzuniga-amd:
Why is it not possible to directly replace the types in the client code?

tfalders:
> Why is it not possible to directly replace the types in the client code?

I was getting a number of compiler errors that I didn't have an easy way to fix. One set of errors was related to being unable to assign a float/double to the hip complex type, which we take advantage of in our test init methods. Another set was caused by ambiguous calls to +=, -=, etc..., which I couldn't figure out how to fix.

cgmb (approved):

jzuniga-amd (approved):

PR #41, opened 2021-08-11
Fix HIP_PATH check

### LINKED ISSUE #40: wrong check if environment variable HIP_PATH is defined
in the root CMakeLists.txt, currently in the develop branch on lines 58--62, the HIP_PATH cmake variable is (supposed to be) set depending on the HIP_PATH environment variable:
```
if( NOT DEFINED $ENV{HIP_PATH})
    set( HIP_PATH "/opt/rocm/hip" )
else( )
    set (HIP_PATH $ENV{HIP_PATH} )
endif( )
```
but this is the wrong way of checking if an environment variable is defined. See
[https://cmake.org/cmake/help/latest/variable/ENV.html](https://cmake.org/cmake/help/latest/variable/ENV.html)
or
[https://cmake.org/pipermail/cmake/2011-October/046706.html](https://cmake.org/pipermail/cmake/2011-October/046706.html)

the correct way is
```
if( NOT DEFINED ENV{HIP_PATH})
    set( HIP_PATH "/opt/rocm/hip" )
else( )
    set (HIP_PATH $ENV{HIP_PATH} )
endif( )
```
(note the missing dollar in the if)

the consequence of this bug is that building hipSOLVER is not as straightforward as executing `cmake ..` and `make` when HIP is installed in a non-default location
The workaround seems to be adding an additional flag `-DCMAKE_MODULE_PATH="${HIP_PATH}/cmake"` to cmake

cgmb:
Fix `HIP_PATH` check in CMakeLists.txt. Thanks to @jakub329homola, for pointing this out.

Fixes #40.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@ Full documentation for hipSOLVER is available at [hipsolver.readthedocs.io](http
 ### Deprecated
 ### Removed
 ### Fixed
+- Fixed use of incorrect `HIP_PATH` when building from source (#40).
+  Thanks [@jakub329homola](https://github.com/jakub329homola)!
+
 ### Known Issues
 ### Security
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,7 +55,7 @@ include( ROCMInstallSymlinks )
 set ( VERSION_STRING "1.2.0" )
 rocm_setup_version( VERSION ${VERSION_STRING} )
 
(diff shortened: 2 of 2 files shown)


daineAMD (approved):
I'll make the same change in hipBLAS for now as well.

tfalders (approved):
Looks good, though CI is failing. Seems to be a rocBLAS/rocSOLVER interface issue.

jzuniga-amd (approved):

PR #42, opened 2021-08-27
Fix hipsolver-test on Windows

cgmb:
This change should fix the hipSOLVER tests on the Windows CI. Targeting master, as this will be a hotfix.

The brunt of the change is copying hipsolver.dll into the clients/staging directory. At some point, we will probably switch to just updating the PATH before running the tests, but copying DLLs is the way the libraries are doing things for now. The DLL copying strategy was chosen because it ensures that the code we're testing is at the top of [the Windows DLL search priority list](https://docs.microsoft.com/en-us/windows/win32/dlls/dynamic-link-library-search-order#standard-search-order-for-desktop-applications).

Issue: SWDEV-301261 (HOTFIX)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,7 +97,7 @@ endif()
 if( USE_CUDA)
     find_package( HIP MODULE REQUIRED )
 else( )
-    find_package( hip REQUIRED PATHS ${HIP_PATH} ${ROCM_PATH} /opt/rocm )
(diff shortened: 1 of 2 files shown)


tfalders (approved):

cgmb:
The proof will be in the pudding when the Windows CI is expanded to start checking hipSOLVER PRs.

cgmb:
The Windows CI has been enabled on hipSOLVER PRs and they are passing.

jzuniga-amd (approved):

cgmb:
I'm going to merge to `release/21.40`, as that does not require any additional approvals.

PR #44, opened 2021-09-22
Changelog update for ROCm 4.5

lawruble13:


tfalders (approved):
This is a bit nitpicky, but I'd prefer if the added functions could be moved to the top of the list rather than the bottom. The current changelog for 4.4 is already a combination of 4.2 and 4.4 with the functions for 4.2 on the bottom. I don't think it makes sense to have the list go 4.4 -> 4.2 -> 4.5.

lawruble13:
Sure, I would agree that doesn't make much sense. They had been put on the bottom so that the natural reading order matched the chronological ordering of the additions, but I wasn't aware that the 4.2 functions were already on the bottom.

jzuniga-amd (approved):

cgmb (approved):

PR #56, opened 2021-10-29
Delete docker directory

cgmb:
The docker directory was used for math-ci, but is no longer needed. See https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/283.


tfalders (approved):
Awesome! I never liked this directory anyway.

jzuniga-amd (approved):

PR #61, opened 2021-11-01
Address cppcheck issues

tfalders:
Should address SWDEV-306987.

I'm not sure if this will fix the supposed syntax error in utility.hpp, but it appears to be a false positive anyway.

### DIFF
--- a/clients/include/complex.hpp
+++ b/clients/include/complex.hpp
@@ -13,15 +13,7 @@ private:
     float x, y;
 
 public:
-#if __cplusplus >= 201103L
     hipsolverComplex() = default;
-#else
-    hipsolverComplex()
-        : x(0)
-        , y(0)
-    {
-    }
-#endif
 
     hipsolverComplex(float r, float i = 0)
         : x(r)
@@ -53,15 +45,7 @@ private:
     double x, y;
 
 public:
-#if __cplusplus >= 201103L
     hipsolverDoubleComplex() = default;
-#else
-    hipsolverDoubleComplex()
-        : x(0)
-        , y(0)
-    {
-    }
-#endif
 
     hipsolverDoubleComplex(double r, double i = 0)
         : x(r)
--- a/clients/include/complex.hpp
+++ b/clients/include/complex.hpp
@@ -16,7 +16,11 @@ public:
 #if __cplusplus >= 201103L
     hipsolverComplex() = default;
 #else
-    hipsolverComplex() {}
+    hipsolverComplex()
+        : x(0)
+        , y(0)
+    {
+    }
 #endif
 
     hipsolverComplex(float r, float i = 0)
@@ -52,7 +56,11 @@ public:
 #if __cplusplus >= 201103L
     hipsolverDoubleComplex() = default;
 #else
-    hipsolverDoubleComplex() {}
+    hipsolverDoubleComplex()
+        : x(0)
+        , y(0)
+    {
+    }
 #endif
 
     hipsolverDoubleComplex(double r, double i = 0)
--- a/clients/include/testing_ormqr_unmqr.hpp
+++ b/clients/include/testing_ormqr_unmqr.hpp
@@ -590,8 +590,9 @@ void testing_ormqr_unmqr(Arguments& argus)
 
     // validate results for rocsolver-test
     // using s * machine_precision as tolerance
+    int s = left ? m : n;
     if(argus.unit_check)
-        ROCSOLVER_TEST_CHECK(T, max_error, (left ? m : n));
+        ROCSOLVER_TEST_CHECK(T, max_error, s);
 
     // output results for rocsolver-bench
     if(argus.timing)
--- a/clients/include/testing_ormqr_unmqr.hpp
+++ b/clients/include/testing_ormqr_unmqr.hpp
@@ -590,9 +590,8 @@ void testing_ormqr_unmqr(Arguments& argus)
 
     // validate results for rocsolver-test
     // using s * machine_precision as tolerance
-    int s = left ? m : n;
     if(argus.unit_check)
-        ROCSOLVER_TEST_CHECK(T, max_error, s);
+        ROCSOLVER_TEST_CHECK(T, max_error, (left ? m : n));
 
     // output results for rocsolver-bench
     if(argus.timing)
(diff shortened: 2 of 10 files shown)


clients/include/complex.hpp
```diff
@@ -16,7 +16,11 @@ typedef struct hipsolverComplex
 #if __cplusplus >= 201103L
     hipsolverComplex() = default;
 #else
-    hipsolverComplex() {}
+    hipsolverComplex()
+        : x(0)
+        , y(0)
+    {
+    }
 #endif
```
cgmb, line 24: Could you do a clean build with `#error hipsolverComplex else branch` in that `#else`? I suspect `#if __cplusplus >= 201103L` is always true. If that's the case, we can just delete this branch of conditional logic entirely.

clients/include/complex.hpp
```diff
@@ -52,7 +56,11 @@ typedef struct hipsolverDoubleComplex
 #if __cplusplus >= 201103L
     hipsolverDoubleComplex() = default;
 #else
-    hipsolverDoubleComplex() {}
+    hipsolverDoubleComplex()
+        : x(0)
+        , y(0)
+    {
+    }
 #endif
```
cgmb, line 64: Ditto. `#error hipsolverDoubleComplex else branch`

clients/include/utility.hpp
```diff
@@ -148,6 +146,9 @@ static constexpr bool is_complex = false;
 #define HIPSOLVER_CLANG_STATIC
 #endif
 
+template <typename T>
+HIPSOLVER_CLANG_STATIC constexpr bool is_complex = false;
+
```
cgmb, line 151: What's the problem?

cgmb: Oh. I see. That's your workaround for the cppcheck syntax error.

I've been having trouble reproducing the scan results myself, too. Nothing here is urgent, so I'd like to ensure we've fully addressed the problem before moving forward. We probably just need to find the right person to ask about reproducing the error locally.

tfalders: Looks like the syntax error is still present in my local results. I guess we either have to suppress the warning or wait for an update to cppcheck?

jzuniga-amd (approved):

cgmb (changes requested):
I ran the latest version of cppcheck on your branch. Paul Fultz has been doing some fixes on cppcheck ([example](https://github.com/danmar/cppcheck/commit/8bff45281dcd8a1794c310255729f2012f750c13)), and I figured if someone from ROCm has been working on it, the fixes are likely to be relevant. The log is attached:  [hipsolver_cppcheck_results.txt](https://github.com/ROCmSoftwarePlatform/hipSOLVER/files/7560596/hipsolver_cppcheck_results.txt)

There are a lot more warnings with this version of cppcheck. Most of them can be ignored, but there are a few worth addressing.

Please fix:
- All unreadVariable, unusedVariable, functionConst and duplicateBreak warnings.
- All passedByValue warnings for `std::string`. (This is not very important in our code, but it's good practice and trivial to do.)
- The shadowVariable warnings in clients/common/utility.cpp.

Please check:
- If the duplicateAssignExpression warning is a problem. I assume it's fine, but please confirm.

Everything else you can ignore. You don't even need to bother suppressing them until we get another report. Some of it is good advice. Some of it's a matter of style. None of it is new or surprising, though. In particular, clients/rocblascommon should basically be re-written anyway. That code needs more than just fixing lint warnings.

tfalders:
I'd forgotten these changes had been requested. Sorry for the wait.

> * If the duplicateAssignExpression warning is a problem. I assume it's fine, but please confirm.

Looks like it's just complaining that two variables in a row are being declared and assigned the same expression. It should be fine to leave it and I like this style for its clarity.

clients/include/testing_ormqr_unmqr.hpp
```diff
@@ -590,9 +590,8 @@ void testing_ormqr_unmqr(Arguments& argus)
 
     // validate results for rocsolver-test
     // using s * machine_precision as tolerance
-    int s = left ? m : n;
     if(argus.unit_check)
-        ROCSOLVER_TEST_CHECK(T, max_error, s);
+        ROCSOLVER_TEST_CHECK(T, max_error, (left ? m : n));
```
cgmb, line 594: What was this one complaining about? It seems like it was better before, where you had a name for `left ? m : n` that you use to refer to the value in the comments.

If it's just complaining that the variable scope could be reduced, maybe put it inside the `if(argus.unit_check)` block instead.

clients/include/testing_ormtr_unmtr.hpp
```diff
@@ -623,9 +623,8 @@ void testing_ormtr_unmtr(Arguments& argus)
 
     // validate results for rocsolver-test
     // using n * machine_precision as tolerance
-    int s = left ? m : n;
     if(argus.unit_check)
-        ROCSOLVER_TEST_CHECK(T, max_error, s);
+        ROCSOLVER_TEST_CHECK(T, max_error, (left ? m : n));
```
cgmb, line 627: Just noticed that `n * machine_precision` should probably have been `s * machine_precision`. Also, same comment as above.

clients/include/testing_ormqr_unmqr.hpp
tfalders: This was one of the unusedVariable warnings. I guess it had trouble sensing that `s` is used in the macro. Perhaps it's time for me to update this to a function like it is in rocSOLVER.

tfalders: Oh, scratch that, I guess it's already the same as in rocSOLVER. I must be thinking of a different macro that was turned into a function.

What do you think? Should we address the unusedVariable warning or ignore it?

cgmb: The way that cppcheck is being invoked is not ideal. We're not giving it the proper include directories to find all the macro definitions. When we integrate it into our CI builds, it will be done better, and hopefully that issue will go away.

cgmb: As a general rule: don't make your code worse just to satisfy a linter. They exist to bring your attention to something that *might* be a problem. If you've taken a careful look at something cppcheck is complaining about, and it's not giving you good advice, then ignore it.

cgmb (approved):

clients/include/testing_ormqr_unmqr.hpp
tfalders: It was only _slightly_ worse, but I agree that this is better. I just need to remember to keep the comments up-to-date.

PR #67, opened 2021-11-30
Install script improvements

tfalders:
Ports some of the nice features of rocSOLVER's install script to hipSOLVER, including the cmake-arg argument and the ability to specify relative paths for custom rocBLAS and rocSOLVER locations.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -10,18 +10,18 @@ function display_help()
 {
   echo "hipSOLVER build & installation helper script"
   echo "./install [-h|--help] "
-  echo "    [-h|--help] prints this help message"
-  echo "    [-i|--install] install after build"
-  echo "    [-d|--dependencies] install build dependencies"
-  echo "    [-c|--clients] build library clients too (combines with -i & -d)"
+  echo "    [-h|--help] Prints this help message"
+  echo "    [-i|--install] Install after build"
+  echo "    [-d|--dependencies] Install build dependencies"
+  echo "    [-c|--clients] Build library clients too (combines with -i & -d)"
   echo "    [-g|--debug] -DCMAKE_BUILD_TYPE=Debug (default is =Release)"
   echo "    [-k|--relwithdebinfo] -DCMAKE_BUILD_TYPE=RelWithDebInfo."
-  echo "    [-r]--relocatable] create a package to support relocatable ROCm"
-  echo "    [--cuda|--use-cuda] build library for cuda backend"
+  echo "    [-r]--relocatable] Create a package to support relocatable ROCm"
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -304,6 +310,8 @@ build_static=false
 build_release_debug=false
 build_codecoverage=false
 build_address_sanitizer=false
+declare -a cmake_common_options
+declare -a cmake_client_options
```
cgmb, line 314: You're declaring these variables as arrays (which is good) and using array append for the --cmake-arg flag (`+=("element"`), but still using string concatenation elsewhere (e.g., `cmake_common_options="${cmake_common_options} element"`). The first string concatenation is going to change the variable type from array to string.

cgmb:
The CI on hipSOLVER is cheap and it does at least run the install script.

install.sh
tfalders: Ah, my ignorance of shell scripting is revealed! I'll update the rest of the cmake_common_options additions.

install.sh
```diff
@@ -462,55 +481,53 @@ pushd .
...
+    cmake_common_options+=("-DUSE_CUDA=ON")
   else
-    cmake_common_options="${cmake_common_options} -DUSE_CUDA=OFF"
+    cmake_common_options+=("-DUSE_CUDA=OFF")
   fi
 
   # clients
   if [[ "${build_clients}" == true ]]; then
-    cmake_client_options="${cmake_client_options} -DBUILD_CLIENTS_SAMPLES=ON -DBUILD_CLIENTS_TESTS=ON -DBUILD_CLIENTS_BENCHMARKS=ON"
+    cmake_client_options+=("-DBUILD_CLIENTS_SAMPLES=ON -DBUILD_CLIENTS_TESTS=ON -DBUILD_CLIENTS_BENCHMARKS=ON")
```
cgmb, line 516: Almost. Take a look at how it was done in rocSOLVER. What you've done works, but the value of using an array is that you can correctly handle arguments that contain spaces. (Not that we do, but it becomes possible.)

When you add multiple arguments separated by spaces in a single array element, you then need to split on spaces later, which breaks the original reason for using the array.

tfalders: I tried to split up this string like in rocSOLVER, but it didn't seem to work properly for me. I can give it another try today.

jzuniga-amd (approved):
LGTM

install.sh
tfalders: Looks like I missed the `[@]` where cmake_common_options and cmake_client_options are used. Separating the array elements with spaces seems to work now.

cgmb (approved):

PR #68, opened 2021-12-08
Added compatibility API

tfalders:
Addresses SWDEV-311789.

This PR adds a compatibility API to hipSOLVER whose method signatures are fully consistent with the cuSOLVER API. Functions in this new API have the prefix `hipsolverDn`.

I've also added tests for those functions with signatures that differ between the regular and compatibility APIs, and added documentation for the compatibility API that is largely just a copy of the regular API documentation with its own separate section in the document.


cgmb (commented):
The alternate names could be included in the ABI using the [alias attribute](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#Common-Function-Attributes). Whether that's worth doing or not depends on what we want from our ABI.

I'd also suggest that if we're rearranging our headers, we should put the internal headers into an `internal` directory like in rocBLAS to make it more obvious that users are not supposed to include them directly. (That's also the purpose/meaning of the term 'detail' in hcc_detail or nvcc_detail.)

cgmb:
(Actually, that reminds me. [HIP](https://github.com/ROCm-Developer-Tools/hipamd/tree/develop/include/hip) moved their headers from hcc_detail and nvcc_detail to amd_detail and nvidia_detail a while back. Perhaps we should do the same? Probably not as part of this change, but it's something for the backlog.)

library/src/hcc_detail/hipsolver_compat.cpp
```diff
@@ -0,0 +1,139 @@
...
+ * ************************************************************************ */
+
+#include "hipsolver.h"
+
+extern "C" {
+
+// gesvd
+hipsolverStatus_t hipsolverDnSgesvd_bufferSize(hipsolverDnHandle_t handle, int m, int n, int* lwork)
+{
+    return hipsolverSgesvd_bufferSize(handle, 'N', 'N', 0, 0, lwork);
```
jzuniga-amd, line 12: Similarly to getrf, I think this helper should return a non-zero value when 'm' and 'n' are not zero. Here the things are more complicated, though, as we don't know the values of `jobu` and `jobv`. I see two options:
a. Get the workspace size for all the 15 possible permutations of `jobu` and `jobv`, and return the maximum.
b. Modify `gesvd_getMemorySize` in rocsolver to always return the maximum regardless of `jobu` and `jobv` values.

I don't think that the difference in size for the different values of `jobu` and `jobv` is substantial if any. So, we could do option b and, eventually, as the values of `jobu` and `jobv` will become useless, we could then remove them altogether from the APIs. But this sound like a longer path, so for now, I would suggest we follow option a. What do you think?

library/src/hcc_detail/hipsolver_compat.cpp
```diff
@@ -0,0 +1,139 @@
...
+                                    int                 ldu,
+                                    float*              V,
+                                    int                 ldv,
+                                    float*              work,
+                                    int                 lwork,
+                                    float*              rwork,
+                                    int*                devInfo)
+{
+    return hipsolverSgesvd(
+        handle, jobu, jobv, m, n, A, lda, S, U, ldu, V, ldv, NULL, 0, rwork, devInfo);
```
jzuniga-amd, line 48: Once the bufferSize helper is corrected, we should be able to pass `work` and `lwork` to `hipsolverXgesvd`, as we do on the Nvidia side; so, this implementation should go into the common file.

jzuniga-amd (changes requested):
Some initial comments are below. I have not finished reviewing all changes.

jzuniga-amd (commented):
I left some comments and suggestions (mainly to the documentation) in a PR to your fork. Please have a look and let me know what you think.

jzuniga-amd (approved):
LGTM!

cgmb (approved):
I do wonder if there might be some drawbacks to calling our externally visible functions from inside our library, but I don't know of any specific problems it would cause.

I have no specific concerns. LGTM. Thanks for getting that put together!

library/include/internal/hipsolver-compat.h
```diff
@@ -0,0 +1,1592 @@
...
+ *  use of functions from hipsolver-functions.h is recommended.
+ */
+
+#ifndef HIPSOLVER_COMPAT_H
+#define HIPSOLVER_COMPAT_H
+
+#include "hipsolver-functions.h"
+#include "hipsolver-types.h"
+
+typedef void* hipsolverDnHandle_t;
```
cgmb: I missed this before, but why do we need `hipsolverDnHandle_t`?

tfalders: The cuSOLVER handle type is called `cusolverDnHandle_t`. I figured since we're going for convenience with the compatibility API, more closely matching the handle type name would be convenient too.

cgmb: My concern is that users may assume that they need a `hipsolverDnHandle_t` to use the compatibility functions. We can explain to them that they do not, but it's better to not have to explain anything at all.

A proliferation of functions is no big deal, but a proliferation of types causes (real or perceived) incompatibilities. Given that we want people to use the original hipsolver API whenever possible, they will probably be using a mix of both APIs in the same code. The fact that the types don't match may be a bit confusing.

jzuniga-amd: That is a good point!
And in that case, we may also eliminate the Dn create, destroy, setstream and getstream helpers. In other words, the compat API will contain only functionality APIs.

tfalders: I don't mind removing the handle type, but I'm a little more dubious when it comes to removing functions (even helper functions).

Actually, would it make more sense to define `hipsolverDnHandle_t` in terms of `hipsolverHandle_t`? That way, the connection will be obvious.

cgmb: > would it make more sense to define `hipsolverDnHandle_t` in terms of `hipsolverHandle_t`? That way, the connection will be obvious.

Maybe we could do that, but use `hipsolverHandle_t` in all our API signatures? Then it becomes very obvious you can use `hipsolverHandle_t` everywhere, but you can still port your code with `s/hipsolver/cusolver/g`.

tfalders: Sounds good. Let's do that.

PR #69, opened 2021-12-10
Fix undefined LAPACK symbols in ASan builds

cgmb:
When `$ROCM_PATH/llvm/bin` is prepended to the `PATH`, our `./install.sh` script is building LAPACK with flang (as the first Fortran compiler in the `PATH`). However, hipSOLVER overrides the CMake default to set its own default to gfortran in the main CMakeLists.txt. Thus, when no Fortran compiler is explicitly specified, each component ends up choosing a different compiler. 

This PR will:
* Ensure dependencies use the same compiler selection logic as hipsolver
* Switch to using an INTERFACE library for AddressSanitizer flags

SWDEV-313529


cgmb:
Related: https://github.com/ROCmSoftwarePlatform/hipBLAS/pull/436.

tfalders (approved):
LGTM!

jzuniga-amd (approved):

PR #72, opened 2022-01-10
Allow gels and gesv to be called in-place

tfalders:
Instead of always calling gels_outofplace and gesv_outofplace from hipSOLVER, we will now call the in-place gels and gesv when B == X.

This also removes the Fortran and compatibility API tests for getrf_npvt, and merges the regular API getrf_npvt test into the tests for getrf.


cgmb (approved):
LGTM

jzuniga-amd (approved):
LGTM!

PR #73, opened 2022-01-10
Allow gesvd's rwork argument to be null

tfalders:
Addresses SWDEV-313829.

Also fixes potential issues with other functions that need expanded workspaces.


jzuniga-amd (approved):
Good catch!
LGTM.

cgmb (commented):
The changes look good, though we should update the changelog. We also forgot that for #72.

cgmb:
👍

PR #74, opened 2022-01-10
Added syevj

tfalders:
Addresses SWDEV-310948 and SWDEV-310950.

Although syevj_batched is supposedly a batched function, in reality it's a strided batched function without the strides. The test clients treat it as if it were a strided batched function. Also, despite being a compatibility-only function, I added the implementations to hipsolver.cpp instead of hipsolver-compat.cpp, since the implementations are more like the ones in hipsolver.cpp.


library/src/hcc_detail/hipsolver.cpp
```diff
@@ -5642,6 +5694,904 @@ catch(...)
...
+                                                                   nullptr,
+                                                                   nullptr,
+                                                                   nullptr));
+    rocblas_stop_device_memory_size_query((rocblas_handle)handle, &sz);
+
+    size_t size_E = 0;
+
+    // space for E array
+    if(n > 0)
+        size_E = sizeof(float) * n;
```
cgmb, line 5733: This is purely style, but I would use the conditional operator for this. The core difference between `if` and `?` is that the latter is an expression that returns a value, so my rule of thumb is to use `?` whenever I need a single value that depends on the result of a condition. I reserve `if` for when I don't need an expression because I'm calling functions without return values, or when there's multiple values that depend on a condition.

    size_t size_E = n > 0 ? sizeof(float) * n : 0;

That said, it's a matter of personal preference. If there's a reason you prefer this way, I'm totally fine with that.

library/src/hcc_detail/hipsolver.cpp
```diff
@@ -5642,6 +5694,904 @@ catch(...)
...
+                                    float*               A,
+                                    int                  lda,
+                                    float*               D,
+                                    float*               work,
+                                    int                  lwork,
+                                    int*                 devInfo,
+                                    hipsolverSyevjInfo_t params)
+try
+{
+    std::unique_ptr<rocblas_device_malloc> mem;
```
cgmb, line 5931: `rocblas_device_malloc` already follows the null object pattern for moved-from instances. We should really just add a default constructor to `rocblas_device_malloc` that creates such a null object. It would avoid all this extra indirection.

library/src/hcc_detail/hipsolver.cpp
```diff
@@ -384,6 +385,57 @@ catch(...)
...
+        return HIPSOLVER_STATUS_INVALID_VALUE;
+
+    delete(int*)info;
+
+    return HIPSOLVER_STATUS_SUCCESS;
+}
+
+hipsolverStatus_t hipsolverDnXsyevjSetMaxSweeps(hipsolverSyevjInfo_t info, int max_sweeps)
+{
+    return HIPSOLVER_STATUS_NOT_SUPPORTED;
```
cgmb, line 412: We should mention this in the docs.

cgmb (commented):
It looks pretty good. I have a few suggestions.

jzuniga-amd (changes requested):
It looks good to me, but:
- I think it would be better to finalize the 'null E' fixes before merging this one.
- Also, as @cgmb mentioned, the docs should clearly indicate that the rocsolver backend is ignoring the jacobi parameters, especially tolerance and number of iterations, ie. the eigenvalues are always computed to machine precision. We should mention that we included these APIs for compatibility but, until the corresponding APIs in rocsolver are ready, behind the scenes, the computations are done with the classic QR algorithm.

jzuniga-amd:
I would add these documentation details in the same section of the userguide intro for now, but I think there are enough reasons now to start documenting hipsolver APIs in the corresponding documentation chapters (will open ticket to work on it next dev cycles).

jzuniga-amd (approved):
LGTM. 
Just be sure to unify the code with the use of the new rocblas_device_malloc. 
(Maybe you can merge this one and then do all changes in PR#75.)

cgmb (approved):
> Just be sure to unify the code with the use of the new rocblas_device_malloc.
> (Maybe you can merge this one and then do all changes in PR#75.)

ditto
(Ditto!)

PR #75, opened 2022-01-13
Workspace fixes

tfalders:
While working on one of the compatibility functions, I discovered an error in the bufferSize methods for those functions that need extra workspace for missing arguments (e.g. gesvd, syevd, sygvd), which can result in some rare allocation errors with automatic workspace management. We need to call `rocblas_set_optimal_device_memory_size` after computing the expanded size since rocBLAS does some rounding up behind the scenes.

I've also added a null handle check to the bufferSize methods to ensure the same error status precedence that we have in rocSOLVER remains in hipSOLVER.

Finally, I did a bit of cleanup in the functions that need extra workspace so that the function call is not duplicated in the code.

This doesn't need to get into the next release, but I will be applying the changes to the functions I'm currently working on.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@ Full documentation for hipSOLVER is available at [hipsolver.readthedocs.io](http
 ### Removed
 ### Fixed
 - bufferSize functions will now return HIPSOLVER_STATUS_NOT_INITIALIZED instead of HIPSOLVER_STATUS_INVALID_VALUE when both handle and lwork are null.
-- Fixed rare memory allocation failure in syevd/heevd and sygvd/hegvd caused by allocating workspace arrays outside of rocSOLVER.
+- Fixed rare memory allocation failure in syevd/heevd and sygvd/hegvd caused by improper workspace array allocation outside of rocSOLVER.
 
 ### Known Issues
 ### Security
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@ Full documentation for hipSOLVER is available at [hipsolver.readthedocs.io](http
 ### Removed
 ### Fixed
 - bufferSize functions will now return HIPSOLVER_STATUS_NOT_INITIALIZED instead of HIPSOLVER_STATUS_INVALID_VALUE when both handle and lwork are null.
-- Fixed rare memory allocation failure in syevd/heevd and sygvd/hegvd when using rocblas_device_malloc to allocate multiple workspace arrays.
+- Fixed rare memory allocation failure in syevd/heevd and sygvd/hegvd caused by allocating workspace arrays outside of rocSOLVER.
 
 ### Known Issues
 ### Security
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ Full documentation for hipSOLVER is available at [hipsolver.readthedocs.io](http
 ### Deprecated
 ### Removed
 ### Fixed
+- Fixed potential workspace allocation failure in some functions.
+
 ### Known Issues
 ### Security
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,8 @@ Full documentation for hipSOLVER is available at [hipsolver.readthedocs.io](http
 ### Deprecated
 ### Removed
 ### Fixed
-- Fixed potential workspace allocation failure in some functions.
+- bufferSize functions will now return HIPSOLVER_STATUS_NOT_INITIALIZED instead of HIPSOLVER_STATUS_INVALID_VALUE when both handle and lwork are null.
+- Fixed rare memory allocation failure in syevd/heevd and sygvd/hegvd when using rocblas_device_malloc to allocate multiple workspace arrays.
 
 ### Known Issues
 ### Security
--- a/library/src/hcc_detail/hipsolver.cpp
+++ b/library/src/hcc_detail/hipsolver.cpp
@@ -17,7 +17,6 @@
 #include <functional>
 #include <iostream>
 #include <math.h>
-#include <memory>
 
 using namespace std;
 
--- a/library/src/hcc_detail/hipsolver.cpp
+++ b/library/src/hcc_detail/hipsolver.cpp
@@ -3249,13 +3249,15 @@ try
                                                                    nullptr));
     rocblas_stop_device_memory_size_query((rocblas_handle)handle, &sz);
 
+    size_t size_E = 0;
+
     // space for E array (aka rwork)
     if(min(m, n) > 0)
-        sz += sizeof(float) * min(m, n);
+        size_E = sizeof(float) * min(m, n);
 
     // update size
     rocblas_start_device_memory_size_query((rocblas_handle)handle);
-    rocblas_set_optimal_device_memory_size((rocblas_handle)handle, sz);
+    rocblas_set_optimal_device_memory_size((rocblas_handle)handle, sz, size_E);
     rocblas_stop_device_memory_size_query((rocblas_handle)handle, &sz);
 
     if(status != HIPSOLVER_STATUS_SUCCESS)
@@ -3301,13 +3303,15 @@ try
                                                                    nullptr));
     rocblas_stop_device_memory_size_query((rocblas_handle)handle, &sz);
 
+    size_t size_E = 0;
+
     // space for E array (aka rwork)
     if(min(m, n) > 0)
-        sz += sizeof(double) * min(m, n);
+        size_E = sizeof(double) * min(m, n);
 
     // update size
     rocblas_start_device_memory_size_query((rocblas_handle)handle);
-    rocblas_set_optimal_device_memory_size((rocblas_handle)handle, sz);
+    rocblas_set_optimal_device_memory_size((rocblas_handle)handle, sz, size_E);
     rocblas_stop_device_memory_size_query((rocblas_handle)handle, &sz);
 
     if(status != HIPSOLVER_STATUS_SUCCESS)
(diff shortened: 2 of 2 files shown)


cgmb (changes requested):
Interesting. It's nice to catch a bug like that. I'm not thrilled by the code, though it's not really your fault. We need better abstractions for dealing with the rocblas workspace.

The use of `std::unique_ptr` with `rocblas_device_malloc` is very clunky. The awkward way you have to use it has already has lead to an error in your implementation. Additionally, I'm reminded that hipSOLVER depends on the rocBLAS internal API for `rocblas_device_malloc`. There's no good reason for that. We really should take the things that we've learned working with the rocBLAS internal memory API over the past year and stabilize a public C API for the functionality.

If we want to squeeze this fix into the next release, I don't necessarily object. However, I think we'd be sacrificing code quality to do it. That's fine to do temporarily, but I think we'd want to be sure we have a plan to rework it into a proper long-term solution.

library/src/hcc_detail/hipsolver.cpp
```diff
@@ -3353,53 +3456,38 @@ try
...
             hipsolverSgesvd_bufferSize((rocblas_handle)handle, jobu, jobv, m, n, &lwork));
         CHECK_ROCBLAS_ERROR(hipsolverManageWorkspace((rocblas_handle)handle, lwork));
 
-        rocblas_device_malloc mem((rocblas_handle)handle, sizeof(float) * min(m, n));
         if(!rwork && min(m, n) > 1)
         {
+            mem = std::make_unique<rocblas_device_malloc>((rocblas_handle)handle,
+                                                          sizeof(float) * min(m, n));
             if(!mem)
                 return HIPSOLVER_STATUS_ALLOC_FAILED;
```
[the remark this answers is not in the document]
tfalders: Good catch! Thanks

library/src/hcc_detail/hipsolver.cpp
```diff
@@ -396,6 +397,8 @@ hipsolverStatus_t hipsolverSorgbr_bufferSize(hipsolverHandle_t   handle,
                                              int*                lwork)
 try
 {
+    if(!handle)
+        return HIPSOLVER_STATUS_NOT_INITIALIZED;
```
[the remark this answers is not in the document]
tfalders: This is only covering the very specific case where both handle and lwork are null, and making it return HIPSOLVER_STATUS_NOT_INITIALIZED rather than HIPSOLVER_STATUS_INVALID_VALUE. A null handle should otherwise be detected when we call the rocSOLVER function in the memory size query.

Unfortunately, the bufferSize methods are not tested right now, and even if they were, the current bad arg test wouldn't have caught this. In the bad arg tests we have, the null handle test only sets the handle to be null.

tfalders:
> We really should take the things that we've learned working with the rocBLAS internal memory API over the past year and stabilize a public C API for the functionality.

I'm curious to hear what kinds of ideas you have for this, but I'm also concerned that it might end up duplicating functionality, not only with the rocBLAS memory API but also with HIP. Regardless, I'm on board with trying to find a cleaner way to manage workspace for hipSOLVER.

jzuniga-amd:
mmm I wonder whether it would be easier to create an internal API in rocsolver to handle these cases (like with gels_outofplace). We could have, for example, the function rocsolver_gesvd_nulle that will still call the same rocsolver_gesvd_template, but before that, it could handle the null pointer E (return the adjusted size in a memory query, or create a non-null E as part of the workspace)... this way we use the C++ internal API that rocBLAS created to use the rocblas_handle memory functions within rocSOLVER as expected, and we keep hipSOLVER's code simple. What do you think?

library/src/hcc_detail/hipsolver.cpp
```diff
@@ -3353,53 +3464,38 @@ try
...
         if(!rwork && min(m, n) > 1)
         {
-            if(!mem)
+            mem = std::make_unique<rocblas_device_malloc>((rocblas_handle)handle,
+                                                          sizeof(float) * min(m, n));
+            if(!mem->operator bool())
                 return HIPSOLVER_STATUS_ALLOC_FAILED;
-            rwork = (float*)mem[0];
+            rwork = (float*)(*mem)[0];
         }
```
cgmb, line 3481: This awkwardness exists because `rocblas_device_malloc` does not support the move semantics you want, so you're using `std::unique_ptr` which does. However, the purpose of `rocblas_device_malloc` is basically the same as `std::unique_ptr`, so we might as well just add the desired move feature to `rocblas_device_malloc`.

jzuniga-amd: I think the idea of the original design was to always initialize the resources when created (RAII), so to avoid this kind of lazy initialization with `unique_ptr` and all that. That's why I say it would be cleaner to deal with null E in rocsolver, but any option is fine by me (I don't know much about these nuances of the language anyways)... if you want to modify the `rocblas_device_malloc` class I think it would be ok; rocblas does not really use it. 

I have a question, though: in hipsolver code, as it is right now, the `rocblas_device_malloc` object `mem` only lives within the `else` statement, so it can be initialized when created (as expected). What is the problem with this approach? why do we want to change it? Is it just because you have to write twice the call to the rocsolver function (once inside the `if` and once inside the `else`)?

cgmb:
> We could have, for example, the function rocsolver_gesvd_nulle that will still call the same rocsolver_gesvd_template, but before that, it could handle the null pointer E (return the adjusted size in a memory query, or create a non-null E as part of the workspace)

I have no strong feelings about this. My `rocblas_device_malloc` suggestions are mostly because I always felt it should work like `std::unique_ptr` and this is the first time it's been relevant.

library/src/hcc_detail/hipsolver.cpp
tfalders: > Is it just because you have to write twice the call to the rocsolver function (once inside the `if` and once inside the `else`)?

Essentially, yes. It hasn't really been a big issue before, but with gesvdj having an additional call to rocblas_geam, these copied function calls are starting to occupy many lines of code and look rather untidy.

I like @cgmb's idea of adding the relevant features to `rocblas_device_malloc` so we can drop the use of `unique_ptr`. Combined with these changes, the workspace code should look a lot more tidy.

tfalders:
> We could have, for example, the function rocsolver_gesvd_nulle that will still call the same rocsolver_gesvd_template, but before that, it could handle the null pointer E (return the adjusted size in a memory query, or create a non-null E as part of the workspace)

It's certainly an option, though one that we should probably consider for the next release cycle rather than this one.

I would personally prefer to stick with the current approach, since adding new functions (for gesvd, syevd, and sygvd) seems a little overkill when we achieve the same thing by allocating some extra workspace. With @cgmb's changes to `rocblas_device_malloc`, I think the main issues with the current approach can be addressed.

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ Full documentation for hipSOLVER is available at [hipsolver.readthedocs.io](http
 ### Deprecated
 ### Removed
 ### Fixed
+- Fixed potential workspace allocation failure in some functions.
```
cgmb, line 22: This is too vague. Please give enough detail so that someone experiencing this failure can tell that it's been fixed.

library/src/hcc_detail/hipsolver.cpp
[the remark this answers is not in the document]
cgmb: I can live with that. However, this is a user-visible change, so it deserves a mention in the changelog.

cgmb (commented):
If you adjust the use of `rocblas_device_malloc`, I think all that's needed are some tweaks for the changelog.

tfalders:
I noticed some test failures (workspace allocation errors) in GESV_INPLACE after updating rocBLAS and rocSOLVER on my local machine. It appears to be because I'm not calling the bufferSize method as in-place in the test. I've decided to take the maximum of the in-place and out-of-place workspace sizes (rather than one or the other) just in case a user makes the same mistake - this is also consistent with the way we handle getrf and getrf_npvt. I've added the change and the same for GELS to this PR.

jzuniga-amd:
Oh I see. I was about to ask why the change...

CHANGELOG.md
```diff
@@ -32,6 +32,9 @@ Full documentation for hipSOLVER is available at [hipsolver.readthedocs.io](http
 ### Deprecated
 ### Removed
 ### Fixed
+- bufferSize functions will now return HIPSOLVER_STATUS_NOT_INITIALIZED instead of HIPSOLVER_STATUS_INVALID_VALUE when both handle and lwork are null.
+- Fixed rare memory allocation failure in syevd/heevd and sygvd/hegvd when using rocblas_device_malloc to allocate multiple workspace arrays.
```
jzuniga-amd, line 36: maybe better not to mention rocblas_device_malloc...?
This is an internal interface and it is not documented/supported anywhere (not for public use).
I would say:
"Fixed rare memory allocation failure in syevd/heevd and sygvd/hegvd when allocating multiple workspace arrays if the E parameter is null"
or something like this...

jzuniga-amd (approved):
LGTM. Just a minor comment.

CHANGELOG.md
```diff
@@ -32,6 +32,9 @@ Full documentation for hipSOLVER is available at [hipsolver.readthedocs.io](http
 ### Deprecated
 ### Removed
 ### Fixed
+- bufferSize functions will now return HIPSOLVER_STATUS_NOT_INITIALIZED instead of HIPSOLVER_STATUS_INVALID_VALUE when both handle and lwork are null.
```
cgmb, line 35: Was the old behaviour somehow wrong? If so, a fix makes sense. I would have thought it was just a change.

cgmb (approved):
I had a couple questions about things that might (or might not) be small improvements, but it looks good! I like it.

CHANGELOG.md
tfalders: It's only wrong insofar as it doesn't follow rocSOLVER error precedence. In rocSOLVER, a null handle is always the first thing we check. I consider it a fix, but I guess it could go either way.

library/src/hcc_detail/hipsolver.cpp
```diff
@@ -2483,38 +2527,37 @@ hipsolverStatus_t hipsolverZZgels_bufferSize(hipsolverHandle_t handle,
...
+                                                                             n,
+                                                                             nrhs,
+                                                                             nullptr,
+                                                                             lda,
+                                                                             nullptr,
+                                                                             ldb,
+                                                                             nullptr,
+                                                                             ldx,
+                                                                             nullptr));
+    rocblas2hip_status(rocsolver_zgels((rocblas_handle)handle,
```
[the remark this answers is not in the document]
tfalders: My understanding is that `gels_outofplace` with `ldx == ldb` should give the same error codes as `gels`.

PR #76, opened 2022-01-15
Added gesvdj

tfalders:
Addresses SWDEV-310942 and SWDEV-310943.

Like syevjBatched, gesvdjBatched is more like one of our strided batched methods than a batched method, and is treated as such by the tests.


jzuniga-amd (changes requested):
It looks good to me. Good job using geam for the transpose!
I would just suggest the same things as for the syevj PR:
- we need to come to a conclusion for the null E treatment first,
- the documentation in the userguide intro should explain that these APIs ignore the Jacobi tolerance and that the singular values  are always computed to max precision 
- Additionally, we should warn the users that there is an overhead for computing the transpose of the right eigenvectors. (this may go away depending on how we end up implementing the jacobi svd in rocsolver, but for now, we should mention it in the docs).

jzuniga-amd (approved):
LGTM.
Just be sure to unify the code with the use of the new rocblas_device_malloc.
(Maybe you can merge this one and then do all changes in PR#75.)

library/include/internal/hipsolver-compat.h
```diff
@@ -18,6 +18,8 @@
 
 typedef void* hipsolverDnHandle_t;
 
+typedef void* hipsolverGesvdjInfo_t;
```
cgmb, line 21: I sort of wondering if we should be using an opaque type rather than `void`. It's too late to consider the change now, I think, but it would ensure `hipsolverDnHandle_t` and `hipsolverGesvdjInfo_t` are distinct types.

I suppose we can always keep that in mind as something to consider doing during a future major version bump.

cgmb (approved):
:+1:

PR #81, opened 2022-01-24
Hotfix/rpm-license

lawruble13:
Summary of proposed changes:
- Change requested by legal to specify the license type in RPM packages.

### DIFF
--- a/library/CMakeLists.txt
+++ b/library/CMakeLists.txt
@@ -1,5 +1,5 @@
 # ########################################################################
-# Copyright 2016-2021 Advanced Micro Devices, Inc.
+# Copyright 2016-2022 Advanced Micro Devices, Inc.
 # ########################################################################
 
 # This is incremented when the ABI to the library changes
@@ -70,6 +70,7 @@ if( NOT USE_CUDA )
 endif( )
 
 set( CPACK_RESOURCE_FILE_LICENSE "${CMAKE_CURRENT_SOURCE_DIR}/../LICENSE.md" )
+set( CPACK_RPM_PACKAGE_LICENSE "MIT" )
 
 if(WIN32)
   set(CPACK_SOURCE_GENERATOR "ZIP")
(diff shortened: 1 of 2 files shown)


tfalders (approved):

library/CMakeLists.txt
```diff
@@ -1,5 +1,5 @@
 # ########################################################################
-# Copyright 2016-2021 Advanced Micro Devices, Inc.
+# Copyright 2016-2022 Advanced Micro Devices, Inc.
```
cgmb, line 2: Should we maybe update the copyright date in `LICENSE.md` too?

cgmb (approved):

library/CMakeLists.txt
lawruble13: Didn't actually intend to make that change, had it in the stashed commit. The other libraries are still at 2021, so for consistency we could revert this, but updating the copyright dates also seems reasonable. Preferred solution?

jzuniga-amd: If you modify a file in the current year, then you update the copyright. That is the policy we adopted, and I think the clang format script takes care of applying it.

jzuniga-amd (approved):

library/CMakeLists.txt
cgmb: > The other libraries are still at 2021

The main reason I suggested this is that hipSOLVER's LICENSE.md still just says 2020.

lawruble13: > If you modify a file in the current year, then you update the copyright.

So should we leave the copyright date in `LICENSE.md` at 2020, as that file wasn't changed?

cgmb: No. Unlike the copyright notices in other files, `LICENSE.md` is supposed to reflect the whole repository. It only says 2020 because we forgot to update it. It should probably say 2020-2021 for ROCm 5.0 and 2020-2022 for ROCm 5.1.

cgmb (approved):
That works for me. We do have changes from 2022, albeit minor ones.

I don't want to approve that change unilaterally, so please wait for @jzuniga-amd to re-approve.

jzuniga-amd (approved):
Looks good to me.

PR #85, opened 2022-02-26
Only add HIP include directories for CUDA

lawruble13:
This properly allows the dependencies to be in a different location. When using the ROCm backend, the appropriate include directories are added by the CMake targets (`roc::rocblas`, `roc::rocsolver`, and `hip::host`).


cgmb (approved):
This looks fine, though we could improve it if SWDEV-243403 has been implemented.

tfalders (approved):

PR #86, opened 2022-03-08
Packages for test and benchmark executables on all supported OSes using CPack.

lawruble13:


jzuniga-amd (commented):
I guess the same comments in 
https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/386
will apply here...

cgmb (approved):

cgmb:
The conflicts are very minor. Should be pretty easy to resolve. LGTM.

jzuniga-amd (approved):
LGTM.
For consistency, if you can add similar changelog entry, and update the -i flag of the install.sh script as you did in rocSOLVER, that would be nice.

tfalders (approved):

PR #93, opened 2022-04-08
Pin dependency versions for building docs

cgmb:
Jinja2 3.1.0 contains breaking changes for our version of Sphinx. We can resolve this by pinning the version to Jinja2 3.0.3.

Unfortunately, the update that broke docs generation released less than 24 hours before ROCm 5.1 was tagged. This change will fix the problem for ROCm 5.2. I was reluctant to pin _everything_ last time this sort of thing happened because pinning can itself sometimes cause problems, but I think it's our best chance at avoiding this problem in the future.


tfalders:
Should we be keeping requirements.txt under source control if it's auto-generated?

cgmb:
> Should we be keeping requirements.txt under source control if it's auto-generated?

Yes. The generated `requirements.txt` is a function of both `requirements.in` and the current state of the PyPI package index. It's a snapshot of how our dependency constraints resolve today, given a particular set of input parameters and whatever packages are available at the time it was generated.

As new versions of our dependencies are added to PyPI, the generated `requirements.txt` may change even with the same `requirements.in`. When we commit a generated `requirements.txt` into source control, we're saving a known-good configuration for later use. On a semi-regular basis, we will need to re-generate `requirements.txt`, verify that the updated dependencies function correctly, and commit the updated file as our new best-known-configuration.

In short: if we generated `requirements.txt` from `requirements.in` each time we did a docs build, we'd end up with exactly the same behaviour as before this PR. Anything not listed in `requirements.in` could be change due to a new release and possibly break our docs build.

jzuniga-amd (approved):

jzuniga-amd (approved):

tfalders (approved):

PR #95, opened 2022-04-21
Set info to zero when info is not referenced by rocSOLVER

tfalders:
Addresses SWDEV-317485.

When a rocSOLVER function does not have an info parameter but hipSOLVER does, hipSOLVER will now set info to zero.

Note: This PR adds a HIP function call to hipSOLVER. So far, I've tried to stick to hipBLAS's design philosophy of eschewing HIP function calls, but this seemed like the most straightforward solution.


cgmb (approved):
This seems reasonable.

You'll need to merge/rebase on develop to get the fixes for the docs build. Unlike the other CI jobs, the ReadTheDocs job builds exactly the commit you've submitted (as opposed to the commit that would be created as a result of merging the PR).

cgmb:
I suppose this should also go in the changelog.

jzuniga-amd (commented):
I would feel more comfortable keeping the library without calls to hip or cuda and sticking to the original design/idea. 

I wonder whether we should implement the `hipsolverZeroInfo()` function as an internal API in rocSOLVER (as we do with `geqrf_ptr_batched`, `gesv_outofplace`, etc), and then call it from hipSOLVER...

What do you think?

tfalders:
> I wonder whether we should implement the `hipsolverZeroInfo()` function as an internal API in rocSOLVER (as we do with `geqrf_ptr_batched`, `gesv_outofplace`, etc), and then call it from hipSOLVER...
> 
> What do you think?

I think it's a little cleaner this way but I'm also happy to continue sticking to the original design. I've also considered using the quick return of an existing function to set info to zero. That way, we wouldn't have to add a new function, but it does carry some extra overhead and would look weird.

I chatted with Daine and he mentioned that he thinks other hip* libraries do have some HIP code in them, so eschewing HIP calls might be something unique to hipBLAS. Plus, we already have HIP as a dependency due to our use of the complex types.

jzuniga-amd:
Ok. Let's leave it this way then... we can change later if problems arise.

jzuniga-amd (approved):

PR #99, opened 2022-04-28
Remove header wrapper on Windows

urmbista:


cgmb (approved):
This one was fine, but I updated the logic just so it would match rocSOLVER.

jzuniga-amd (approved):

PR #102, opened 2022-05-21
Fix RHEL 8 and 9 client dependencies

cgmb:
When `"$VERSION_ID" == "8.6"` or `"$VERSION_ID" == "9"`, hipSOLVER should install the latest CentOS / RHEL dependencies. Additionally, CMake is available as `cmake` on RHEL 8 and later.

Ticket: SWDEV-338310

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -214,7 +214,7 @@ install_packages( )
 #     yum -y update brings *all* installed packages up to date
 #     without seeking user approval
 #     elevate_if_not_root yum -y update
-      if [[ "${VERSION_ID}" == "8" ]]; then
+      if (( "${VERSION_ID%%.*}" >= "8" )); then
         install_yum_packages "${library_dependencies_centos8[@]}"
         if [[ "${build_clients}" == true ]]; then
           install_yum_packages "${client_dependencies_centos8[@]}"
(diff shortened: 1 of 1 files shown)


cgmb:
See also https://github.com/ROCmSoftwarePlatform/hipBLAS/pull/486.

install.sh
```diff
@@ -214,7 +214,7 @@ install_packages( )
 #     yum -y update brings *all* installed packages up to date
 #     without seeking user approval
 #     elevate_if_not_root yum -y update
-      if [[ "${VERSION_ID}" == "8" ]]; then
+      if (( "${VERSION_ID%%.*}" >= "8" )); then
```
cgmb, line 217: A couple notes on how this works. `${VERSION_ID%%.*}` is a [variable expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html) for `VERSION_ID` that trims all contents matching the wildcard `.*` from the end of the variable. Note: that's a [glob pattern](https://www.man7.org/linux/man-pages/man7/glob.7.html), not a regex. The `(( ))` allows for [integer arithmetic comparisons](https://www.gnu.org/software/bash/manual/html_node/Shell-Arithmetic.html), rather than string comparisons.

jzuniga-amd (approved):

tfalders (approved):

PR #103, opened 2022-05-25
Added syevdx

tfalders:
Adds syevdx to hipSOLVER.

I've also changed the name of argument `D` to `W` in several cases to be more consistent with cuSOLVER.


library/src/hipsolver_module.f90
```diff
@@ -2957,16 +2963,217 @@ function hipsolverZheevd(handle, jobz, uplo, n, A, lda, D, work, lwork, info) &
...
             type(c_ptr), value :: work
             integer(c_int), value :: lwork
             type(c_ptr), value :: info
         end function hipsolverZheevd
     end interface
 
+    ! ******************** SYEVDX/HEEVDX ********************
+    interface
+        function hipsolverSsyevdx_bufferSize(handle, jobz, range, uplo, n, A, lda, vl, vu, il, iu, nev, W, lwork) &
+                result(c_int) &
```
cgmb, line 2976: We've been declaring our return type incorrectly in the Fortran bindings. The `result` declaration takes a variable name, not a type. The right way of doing this would be `result(status)` and then below doing `integer(int_c), value :: status` to the argument list. Or, well, almost. The return value isn't necessarily `c_int` anyway. The type should really be `integer(kind(HIPSOLVER_STATUS_SUCCESS))`. There is also an alternative syntax for the return type that doesn't use `result` at all. You'd just put `integer(kind(HIPSOLVER_STATUS_SUCCESS))` before `function`.

Fixing this one instance would be rather pointless, so I don't think you need to change anything here. However, I don't want to let it go without comment. We really should create a concrete plan to address this issue.

I think it's pretty clear that we're not Fortran experts. I wonder if we might now be able to deprecate this interface so you don't need to keep updating it. The hipFORT project provides these bindings too, and it's maintained by developers with a better understanding of Fortran.

cgmb (approved):

library/src/hipsolver_module.f90
tfalders: Yeah, I'm inclined to let hipFORT take over the Fortran bindings for us. We might want to discuss it in our upcoming meeting. I wish we had a sense for if anyone is using these Fortran bindings right now.

cgmb: I've opened https://github.com/ROCmSoftwarePlatform/hipSOLVER/pull/105 to fix the existing API functions. Could you update the added syevdx functions to match?

tfalders: Sure, once the other PR is merged in.

jzuniga-amd (commented):
Looks good!
Just one concern: we do have an API for syevdx in rocSOLVER but it is only internal (syevdx_inplace). To the public there is no rocsolver_syevdx yet, so I wonder whether we should add only the compatibility API to hipSOLVER for now (i.e. hipsolverDn_syevdx) --as we do with the jacobi methods for example. Thoughts?

clients/benchmarks/client.cpp
```diff
@@ -287,6 +287,40 @@ try
...
+            "                           Used in partial eigenvalue decomposition functions.\n"
+            "                           ")
+
+        ("iu",
+         value<rocblas_int>(),
+            "Upper index in ordered subset of eigenvalues.\n"
+            "                           Used in partial eigenvalue decomposition functions.\n"
+            "                           ")
+
+        ("range",
```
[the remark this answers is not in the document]
jzuniga-amd: mmm I see. Maybe worth to change back rocsolver's erange and eorder to range and order? I have no preference. Not important really, just something to think about.

jzuniga-amd (approved):
Great! That looks good to me.

PR #104, opened 2022-05-25
Add cuda-path to install options

tfalders:


install.sh
```diff
@@ -18,6 +18,7 @@ function display_help()
   echo "    [-k|--relwithdebinfo] -DCMAKE_BUILD_TYPE=RelWithDebInfo."
   echo "    [-r]--relocatable] Create a package to support relocatable ROCm"
   echo "    [--cuda|--use-cuda] Build library for cuda backend"
+  echo "    [--cuda-path] Set specific path to custom built cuda"
```
cgmb, line 21: It's not the style I prefer, but we should use `--cudapath` for consistency with hipBLAS. Many people who build one library or the other from source will likely end up building both, and they'll probably want to copy/paste the build command they used.

cgmb (commented):
This is porting over functionality from https://github.com/ROCmSoftwarePlatform/hipBLAS/pull/475.

install.sh
tfalders: Yeah, I think you might be right. Consistency with hipBLAS is probably more important than consistency with our other option names, though it pains me.

cgmb (approved):

jzuniga-amd (approved):

PR #105, opened 2022-05-30
Fix Fortran result values

cgmb:
The `result` declaration in Fortran defines the name of the result variable, not the type. The result type is not necessarily `c_int`, either. The value returned is an enum.

The hipfort bindings for hipsolver handles this correctly, so this this change is to define the hipsolver API return types in the same way that hipfort does.

### DIFF
--- a/library/src/hipsolver_module.f90
+++ b/library/src/hipsolver_module.f90
@@ -55,7 +55,6 @@ end module hipsolver_enums
 
 module hipsolver
     use iso_c_binding
-    use hipsolver_enums
 
     !---------!
     !   Aux   !
--- a/library/src/hipsolver_module.f90
+++ b/library/src/hipsolver_module.f90
@@ -55,6 +55,7 @@ end module hipsolver_enums
 
 module hipsolver
     use iso_c_binding
+    use hipsolver_enums
 
     !---------!
     !   Aux   !
@@ -64,8 +65,9 @@ module hipsolver
         function hipsolverCreate(handle) &
                 bind(c, name = 'hipsolverCreate')
             use iso_c_binding
+            use hipsolver_enums
             implicit none
-            integer(kind(HIPSOLVER_STATUS_SUCCESS)) :: hipsolverCreate_
+            integer(kind(HIPSOLVER_STATUS_SUCCESS)) :: hipsolverCreate
             type(c_ptr), value :: handle
         end function hipsolverCreate
     end interface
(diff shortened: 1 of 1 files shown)


library/src/hipsolver_module.f90
```diff
@@ -62,41 +62,41 @@ module hipsolver
     
     interface
         function hipsolverCreate(handle) &
-                result(c_int) &
                 bind(c, name = 'hipsolverCreate')
             use iso_c_binding
             implicit none
+            integer(kind(HIPSOLVER_STATUS_SUCCESS)) :: hipsolverCreate_
```
EdDAzevedo, line 68: Have you tried to compile the F90 code with other compilers such as gfortran, just to check the syntax? For example, on line 68, there is reference to HIPSOLVER_STATUS_SUCCESS in another module hipsolver_enums. Perhaps there should be "use hipsolver_enums"? 

The "result" names the variable so we need another line to declare the type of this result variable.

cgmb:
Well, I clearly did this wrong. That's what I get for trying to be lazy about this. I bought a Fortran book. I suppose this afternoon I'll read it. Converting to draft.

EdDAzevedo (approved):

tfalders (approved):
Thanks for taking care of this, @cgmb!

jzuniga-amd (approved):

cgmb:
> The result type is not necessarily c_int, either. The value returned is an enum.

@EdDAzevedo discussed this a bit. I figured I'd share some of what I've learned.

GCC defines the compatible type for enums in [Structures, Unions, Enumerations, and Bit-fields](https://gcc.gnu.org/onlinedocs/gcc/Structures-unions-enumerations-and-bit-fields-implementation.html):

> Normally, the type is `unsigned int` if there are no negative values in the enumeration, otherwise `int`. If `-fshort-enums` is specified, then if there are negative values it is the first of `signed char`, `short` and `int` that can represent all the values, otherwise it is the first of `unsigned char`, `unsigned short` and `unsigned int` that can represent all the values.

This means that hipSOLVER enums on Linux are compatible with `unsigned int`. Fortran does not have unsigned integers, but on all supported platforms, `int` and `unsigned int` are bitwise-identical in the range `[0, INT_MAX]`. So, there should be no problem as long as we don't use any enum values exceeding that.

On Windows, it seems that enums are of type `int`. The MSVC guide on [C enumeration declarations](https://docs.microsoft.com/en-us/cpp/c-language/c-enumeration-declarations?view=msvc-170) states:

> In ANSI C, the expressions that define the value of an enumerator constant always have int type. That means the storage associated with an enumeration variable is the storage required for a single int value.

I don't think any of this affects anything of importance. Just leaving it here for posterity.

cgmb:
It will be a while before the CI is in a passing state.

PR #108, opened 2022-06-13
Remove unnecessary headers

tfalders:
Change suggested by @cgmb. Should address SWDEV-340918.


jzuniga-amd (approved):

cgmb (approved):
Oh. That's nice. I was figured the headers could be trimmed, but I didn't realize they could be removed entirely.  I suppose hipSOLVER is just forwarding from `hipsolver.h` to `cusolverDn.h` so those headers will inherently have to declare the types that we're using here.

tfalders:
This only affects cuda, which is passing, so I'm going to go ahead and force the merge.

PR #109, opened 2022-06-15
Build deps as user

tfalders:
Also removes boost as a dependency and build step.


install.sh
```diff
@@ -481,9 +481,9 @@ if [[ "${install_dependencies}" == true ]]; then
   pushd .
     printf "\033[32mBuilding \033[33mgoogletest & lapack\033[32m from source; installing into \033[33m/usr/local\033[0m\n"
     mkdir -p ${build_dir}/deps && cd ${build_dir}/deps
-    ${cmake_executable} -DCMAKE_INSTALL_PREFIX=deps-install -DBUILD_BOOST=OFF ../../deps
-    make -j$(nproc)
-    make install
+    ${cmake_executable} -DCMAKE_INSTALL_PREFIX=deps-install ../../deps
+    make build_deps
```
cgmb, line 485: This is so slow compared to my solution in https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/438.

deps/CMakeLists.txt
```diff
@@ -74,9 +68,14 @@ if( POLICY CMP0037 )
...
+add_custom_target( build_deps
   ${gtest_custom_target}
   ${lapack_custom_target}
   DEPENDS ${hipsolver_dependencies}
 )
+
+add_custom_target( install_deps
+  ${gtest_install_target}
+  ${lapack_install_target}
+  DEPENDS ${build_deps}
```
cgmb, line 80: I didn't notice when I reviewed the rocblas code, but this doesn't make sense. `build_deps` is not a variable. It's a target name. I think `${build_deps}` evaluates to the empty string.

install.sh
```diff
@@ -481,9 +481,9 @@ if [[ "${install_dependencies}" == true ]]; then
   pushd .
     printf "\033[32mBuilding \033[33mgoogletest & lapack\033[32m from source; installing into \033[33m/usr/local\033[0m\n"
```
cgmb, line 482: This status message is wrong.

docs/source/userguide_install.rst
```diff
@@ -78,7 +78,7 @@ the ``CMAKE_PREFIX_PATH`` definition. The following is a sequence of steps to bu
 
     mkdir -p <HIPSOLVER_BUILD_DIR_PATH>/release/deps
     cd <HIPSOLVER_BUILD_DIR_PATH>/release/deps
-    cmake -DBUILD_BOOST=OFF <HIPSOLVER_SOURCE_PATH>/deps   #assuming boost is installed through package manager as above
+    cmake <HIPSOLVER_SOURCE_PATH>/deps
```
cgmb, line 81: Good catch. I suspect I forgot the docs on rocSOLVER.

cgmb (commented):
I made this exact change for rocSOLVER, then decided I wasn't satisfied with it and created https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/438 instead.

tfalders:
> I made this exact change for rocSOLVER, then decided I wasn't satisfied with it and created [ROCmSoftwarePlatform/rocSOLVER#438](https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/438) instead.

I mostly made this change to help me build on a machine with no admin access and the dependencies not installed, so I'm happy to cancel it if you think there's a better approach. One thing that would be nice to have eventually is the ability to build hipSOLVER (rocSOLVER too) and its dependencies without sudo access. Maybe in that case we only install the dependencies if the `-i` flag is passed.

cgmb:
I didn't notice before but hipSOLVER _already_ builds its deps as user. rocBLAS and rocSOLVER had this problem because they need sudo to install their dependencies into `/usr/local`, but hipSOLVER installs its dependencies into `build/deps/deps-install` and thus doesn't use sudo.

In [rocSOLVER's install.sh](https://github.com/ROCmSoftwarePlatform/rocSOLVER/blob/53074d63a79fa6b9f36acff30dab7bb433cf0ef5/install.sh#L499), the build is done with `elevate_if_not_root make install`.
In [hipSOLVER's install.sh](https://github.com/ROCmSoftwarePlatform/hipSOLVER/blob/6990a56a08c71575d804772f4a02cd044c24ab1e/install.sh#L486), the build is done with `make install`.

> One thing that would be nice to have eventually is the ability to build hipSOLVER (rocSOLVER too) and its dependencies without sudo access.

You should definitely give Spack a try. It serves exactly that purpose and does the job better than anything we might put in our install script.

```
sudo apt install git python3 build-essential gfortran cmake
git clone -b v0.18.0 https://github.com/spack/spack.git
. spack/share/spack/setup-env.sh
spack external find cmake
spack install fmt netlib-lapack googletest
spack load fmt netlib-lapack googletest
./install.sh -c
```

tfalders:
I'm closing this PR because it sounds like it's not necessary and an alternative solution would be desired anyway.

PR #111, opened 2022-06-20
Added sygvdx

tfalders:
Adds sygvdx to hipSOLVER. Like with syevdx, this is going into the compatibility API only, pending an implementation of divide-and-conquer partial eigensolvers.


jzuniga-amd (approved):
LGTM

cgmb (approved):

PR #118, opened 2022-07-25
Add const qualification for compatibility with cuSolver

qjojo:
This adds const qualifiers to the arguments of the hipsolverDn* functions to make the headers more interchangeable with the corresponding cuSolver ones. The const qualification gets removed by const_cast<> once inside hipSolver to impose a limit on the scope of this change, fully propagating everything down all the way wouldn't improve the correctness of hipSolver code and would be a massive undertaking, especially since this is meant as a hotfix.

As a side effect of having run clang-format manually, some functions have been rearranged within the headers as well.


qjojo:
The relevant cuSolver header is cuSolverDn.h

cgmb (commented):
> fully propagating everything down all the way wouldn't improve the correctness of hipSolver code

If you aren't fully propagating it, then the compiler can't verify the correctness. It's true that if all the called functions are effectively `const` despite not being marked as such, then propagating the `const` won't meaningfully change anything. The value is entirely in catching mistakes.

However, if the compiler is not verifying the correctness, then the onus is instead put upon us as developers to do so. If verifying the correctness via inspection is the simplest method, that's fine. Somebody just has to take on that responsibility and do that work.

qjojo:
@cgmb True, but since this is meant as a compatibility hotfix for master and not new development, I think it's probably best to do this for now, and then we can look at applying const qualification throughout the whole codebase for a future version.

tfalders (approved):
This is a good first step and should address the immediate issue. That said, I would like us to get the change fully propagated to the regular API as another hotfix, so that we don't introduce a divergence between the regular and compatibility APIs that lasts a whole release. Plus, most of the `consts` are on pointer arguments to the bufferSize methods, which are all ignored on the rocSOLVER side, meaning most of the `const_casts` won't be necessary.

jzuniga-amd (approved):
LGTM. I agree with Troy.

cgmb (approved):
If Juan and Troy are confident the implementations of these functions are indeed `const`, then that's good enough for me.

> I would like us to get the change fully propagated to the regular API as another hotfix, so that we don't introduce a divergence between the regular and compatibility APIs that lasts a whole release.

I also agree with this. There's only a handful of compatibility API functions that are actually different from the original hipSOLVER APIs. It would be good to preserve that property.

tfalders:
I will force the merge, since the CI failure seems unrelated to this PR.

PR #119, opened 2022-07-28
Various fixes for rectangular gesvdj

tfalders:
This PR hotfixes various issues with gesvdj, mostly related to non-square matrices and the economy size option.

1. Testing code fixes: Under the current testing code when economy size is enabled, V will be allocated as a `min(m,n)*n` matrix with `ldv >= min(m,n)` and U is allocated as `m*m`. This is incorrect. Since gesvdj returns V and not V^H (unlike gesvd), it should actually be a `n*min(m,n)` matrix with `ldv >= n`. Also, we can allocate U as a `m*min(m,n)` matrix instead to save memory.

2. gesvdj_bufferSize fixes: Current code will cause bufferSize to return an error if `ldv < n`, even if V is not required. Also, it's allocating too much workspace by passing `ldv >= n` to rocsolver_gesvd instead of `ldv >= min(m,n)`. I'm cutting down on the amount of workspace needed when economy size is enabled by using rocsolver_gesvd's overwrite functionality, meaning V^H will overwrite A instead of a temporary workspace array, which we can then transpose using rocblas_geam.

3. gesvdj fixes: I mistakenly used `rocblas_operation_none` for the second matrix op in the rocblas_geam calls (figuring it would be ignored), which causes gesvdj to return an error when `m < n`. Other changes to gesvdj are designed to use the overwrite functionality when economy size is enabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,9 +25,12 @@ Full documentation for hipSOLVER is available at [hipsolver.readthedocs.io](http
 
 ### Changed
 - The rocSOLVER backend will now set `info` to zero if rocSOLVER does not reference `info`. (Applies to orgbr/ungbr, orgqr/ungqr, orgtr/ungtr, ormqr/unmqr, ormtr/unmtr, gebrd, geqrf, getrs, potrs, and sytrd/hetrd).
+- gesvdj will no longer require extra workspace to transpose `V` when `jobz` is `HIPSOLVER_EIG_MODE_VECTOR` and `econ` is 1.
 
 ### Fixed
 - Fixed Fortran return value declarations within hipsolver_module.f90
+- Fixed gesvdj_bufferSize returning `HIPSOLVER_STATUS_INVALID_VALUE` when `jobz` is `HIPSOLVER_EIG_MODE_NOVECTOR` and 1 <= `ldv` < `n`
+- Fixed gesvdj returning `HIPSOLVER_STATUS_INVALID_VALUE` when `jobz` is `HIPSOLVER_EIG_MODE_VECTOR`, `econ` is 1, and `m` < `n`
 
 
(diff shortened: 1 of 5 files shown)


jzuniga-amd (commented):
It looks good Troy!
I only found something that seemed weird to me; I don't remember why we left it the way it is, or maybe it is an error that we should correct:
Why do we add the size of array E (or rwork) to the internal workspace?
E/rwork is user provided (in both libraries, cuSOLVER and rocSOLVER); if it is not provided rocSOLVER returns an error and hipSOLVER should too! 
I don't  see why we are managing E as part of the internal workspace. Have we discussed this before?

tfalders:
> E/rwork is user provided (in both libraries, cuSOLVER and rocSOLVER)

I'm not sure what you mean. E/rwork is provided as an argument to cuSOLVER's version of gesvd, but not to its version of gesvdj.

jzuniga-amd:
Yeah, sorry I forgot to mention, this is for gesvd! (I just noticed reviewing this PR :) 

If E/rwork is user provided in cuSOLVER and rocSOLVER, why did we decide to manage E as part of the internal workspace on the rocSOLVER side?

tfalders:
> Yeah, sorry I forgot to mention, this is for gesvd! (I just noticed it reviewing this PR :)

Ah, I see! IIRC, we used to have gesvd always using the user provided rwork, but then a defect was reported because apparently cuSOLVER allows rwork to be null. I had to ensure that we always have enough extra workspace for rwork in case it is provided as null.

jzuniga-amd:
Oh yes, I think I remember now. Sounds good then.

jzuniga-amd (approved):
LGTM

cgmb (approved):
I won't be able to review this in detail, but it looks reasonable.

qjojo:
LGTM

qjojo (approved):

PR #120, opened 2022-08-03
Copyright update

qjojo:
Update copyright messages to be compliant with other ROCm libraries.


tfalders (approved):

clients/benchmarks/client.cpp
```diff
@@ -1,5 +1,24 @@
 /* ************************************************************************
- * Copyright 2020-2022 Advanced Micro Devices, Inc.
+ * Copyright 2020-2022 Advanced Micro Devices, Inc. All rights reserved.
```
cgmb, line 2: Shouldn't there be a `(C)` in here somewhere?

rtest.py
```diff
@@ -1,5 +1,23 @@
...
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IM-
+PLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN CONNE-
+CTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
 Run tests on build"""
```
cgmb, line 21: I know that's how it was before... but why are we putting the license in the file's docstring? I haven't seen any other projects that put the copyright information in the docstring. I only did a quick survey of other projects, but thus far I've only found the license info in comments (e.g., [PySide](https://code.qt.io/cgit/pyside/pyside-setup.git/tree/sources/pyside-tools/project.py)).

qjojo: I think it's just an ergonomics thing, since python doesn't have multi line comments and this saves putting a # on each line. The [top few google results](https://stackoverflow.com/questions/7696924/how-do-i-create-multiline-comments-in-python) mention doing multi line comments this way.

cgmb: That's true. From the BDFL himself:

> Python tip: You can use multi-line strings as multi-line comments. Unless used as docstrings, they generate no code! :-)

This is, however, the file's docstring. The contents of that string will be loaded into the `__doc__` variable at runtime and can be printed with `help('rtest')`.

Does it matter? No, not really. We can do what we want and it's just a minor stylistic matter.

See also [PEP 257 - Docstring Conventions](https://peps.python.org/pep-0257/), which suggests a style convention for docstrings in scripts... but we're not going to put enough time into these scripts to really follow that to the letter anyway.

jzuniga-amd (approved):

qjojo:
Closing in favor of #121

PR #121, opened 2022-08-04
Update copyright information

qjojo:
Same as #120 but for master


tfalders:
If this is the same as #120, then I propose we close #120 and merge this instead. Then we merge master into develop.

jzuniga-amd:
Yes, if this is a hotfix required for 5.3 (master), then this is the expected procedure.

qjojo:
Reopening since we are going forward with this over #120

cgmb:
Note that you can always change the target branch of a PR using the Edit button. Pretty much the only thing you can't change is the source branch.

jzuniga-amd:
If this is a requirement from legal, I think some inconsistencies should be addressed/discussed:
1. I guess the Copyright legend is required to be added to all "code" files (cpp, hpp, etc.) So, why are `potrs_gtest.cpp`, `syevd_heevd_gtest.cpp`, `error_macros.hpp`, and `exceptions.hpp` not updated?
2. What about Fortran code? `hipsolver_module.f90` and `hipsolver_fortran.f90` are not updated either.
3. The python scripts `rmake.py` and `rtest.py` are only for users to build/install the library. If we update them, shouldn't we add the copyright notice to the `install.sh` script as well?

qjojo:
Added the missed fortran files and headers @jzuniga-amd mentioned. I left the comments out of install.sh since that's how rocBLAS is right now: comments in C++, python, and Fortran files but not the install.sh script.

jzuniga-amd:
`clients/gtest/syevd_heevd_gtest.cpp` and `library/src/include/exceptions.hpp` are still missing.

Is there any substantial reason to not update `install.sh` (other than that other libraries have not)? To me, one of the important parts of the copyright legend is the "software is provided as is..." `install.sh` provides software to assist users in the installation process (the same as `rmake.py`) so I think is important it has the legend. 

For the same reasons, I think we should also update the sample codes in `clients/samples/example_basic.c` and `clients/samples/example_basic.cpp`

qjojo:
I just went ahead and added the message to install.sh as well, I think @jzuniga-amd 's reasoning makes sense for including it. I also added it to our CMake scripts since rocBLAS does that as well, they are code after all.

jzuniga-amd (approved):
LGTM.
If we are updating the Cmake files as well, you need to add the main one too `./CMakeLists.txt`

clients/CMakeLists.txt
```diff
@@ -1,5 +1,23 @@
 # ########################################################################
-# Copyright 2016-2022 Advanced Micro Devices, Inc.
+# Copyright 2016-2022 Advanced Micro Devices, Inc. All rights reserved.
```
cgmb, line 2: Doesn't the template use `Copyright (C)`?

cgmb (changes requested):

tfalders (approved):

clients/CMakeLists.txt
qjojo: It does... I just noticed that clang-format is removing it. Is that intentional?

qjojo: Sorry, not clang format. The perl snippet in the pre-commit script that manages copyright dates didn't have a (C) in the replacement string so it was disappearing during commit.

`/usr/bin/perl -pi -e 'INIT { exit 1 if !-f $ARGV[0] || -B $ARGV[0]; $year = (localtime)[5] + 1900 }
            s/^([*\/#[:space:]]*)Copyright\s+(?:\(C\)\s*)?(\d+)(?:\s*-\s*\d+)?/qq($1Copyright $2@{[$year != $2 ? "-$year" : ""]})/ie
            if $. < 10' "$file" && git add -u "$file"`

Fixed so it won't do that anymore.

`/usr/bin/perl -pi -e 'INIT { exit 1 if !-f $ARGV[0] || -B $ARGV[0]; $year = (localtime)[5] + 1900 }
            s/^([*\/#[:space:]]*)Copyright\s+(?:\(C\)\s*)?(\d+)(?:\s*-\s*\d+)?/qq($1Copyright \(C\) $2@{[$year != $2 ? "-$year" : ""]})/ie
            if $. < 10' "$file" && git add -u "$file"`

cgmb: Correct. Updating that script is a necessary part of this change.

cgmb:
Your updated change looks good in principle, but I'm not going to have a chance to go through it in detail. As such, I've dismissed my review.

PR #122, opened 2022-08-04
Add const qualification for compatibility with cuSolver

qjojo:
Same as #118 but on develop branch


tfalders:
Now that #118 is merged into master, it's better if we merge master back into develop. I suggest either revising this PR so that it's a merge from master to develop, or to open a new PR that does that and close this one.

PR #125, opened 2022-08-09
Merge hotfix from #121, #118 into develop

qjojo:
This merges the updated copyright messages on master into develop. It also includes the changes from #118 that modify the const qualification of some compatibility functions, since #124 to merge those into develop has not been approved yet and this is just a merge of master into develop. If this is approved I will close #124, or we can merge #124 first and then this PR.


tfalders:
I only see the changes to const in the commit history. For some reason the updated copyright does not appear to be in here.

qjojo:
> I only see the changes to const in the commit history. For some reason the updated copyright does not appear to be in here.

My mistake, hadn't fetched latest upstream locally. Fixed.

PR #126, opened 2022-08-17
Merge hotfix from #119 into develop

tfalders:
This PR will merge the changes from #119 into develop.

It looks like #118 and #121 are in the commit log even though they were merged in #125, but that's probably because #125 was squashed-and-merged. For future merges from master to develop we might want to consider not squashing the merge.


tfalders:
Looks like this didn't work properly. I'll close and try again.

PR #127, opened 2022-08-17
Merge hotfix from #119 into develop

tfalders:
Corrected version of #126

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,9 +36,12 @@ Full documentation for hipSOLVER is available at [hipsolver.readthedocs.io](http
 
 ### Changed
 - The rocSOLVER backend will now set `info` to zero if rocSOLVER does not reference `info`. (Applies to orgbr/ungbr, orgqr/ungqr, orgtr/ungtr, ormqr/unmqr, ormtr/unmtr, gebrd, geqrf, getrs, potrs, and sytrd/hetrd).
+- gesvdj will no longer require extra workspace to transpose `V` when `jobz` is `HIPSOLVER_EIG_MODE_VECTOR` and `econ` is 1.
 
 ### Fixed
 - Fixed Fortran return value declarations within hipsolver_module.f90
+- Fixed gesvdj_bufferSize returning `HIPSOLVER_STATUS_INVALID_VALUE` when `jobz` is `HIPSOLVER_EIG_MODE_NOVECTOR` and 1 <= `ldv` < `n`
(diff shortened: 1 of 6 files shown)


tfalders:
Apparently the format check is failing, and this is also true on the master branch as well.

What do you guys think we should do? Open a PR against master to fix the format, and then merge into here to fix develop as well? Or just apply the fix in this PR?

jzuniga-amd:
Apply the fix in this PR will only fix develop. If master format is being checked by CQE/mainline guys then we should fix it as well. If not, we can maybe just fix develop (with this PR) for now and then wait until first merge of develop into master (after branching) to get master fixed... 
--Then we should anyways fix 5.3 branch if we don't want to release with "ugly" format :) but that may be easier ?

tfalders:
Hmm... I did want to apply the const changes to the rest of the library as a master hotfix too. Maybe we can do that and fix the format at the same time. @qjojo, would you like me to take care of that? I know you're busy with other things as well.

cgmb:
> If master format is being checked by CQE/mainline guys then we should fix it as well.

I'm pretty sure it's just the math-ci that checks the formatting. That said, I support @tfalders' suggestion to fix it anyway. It's distracting to have checks failing on the release branch, even if they're not particuilarly important.

cgmb:
When you merge back from master to develop, please use "Create a merge commit". The "squash and merge" option creates a new commit with a copy of the content. When you do that for commits that already exist in the official history of the project, it becomes more difficult to see what versions of the software have a particular change.

e.g., You can see that 063d9c12a973c780a9b2c06d51786ec3b8b498ca (Fix reorg warnings) is on the develop branch and available in ROCm 5.2.0 through ROCm 5.2.3.
fix-reorg-warnings

However, https://github.com/ROCmSoftwarePlatform/rocSOLVER/commit/a30e439bec15c58ba7b5e741261640cf923c8d49 (Fix compatibility with older versions of LAPACK) is not shown to be on the develop branch, since the content was copied into a new commit and the relationship with the original change was thereby lost. For the same reason, it will show up as being in ROCm 5.3 when that's tagged, but not ROCm 5.4 or later.
fix-compatibility-with-older-versions-of-lapack

It's not a huge deal, but it's nice to maintain that metadata.

qjojo (approved):

jzuniga-amd (approved):

tfalders:
I'm going to force the merge, and do it as a merge commit so that the PR commits show up in the history. The CI failures appear to be spurious failures.

PR #136, opened 2022-10-27
Remove RPATH/RUNPATH from ROCm libraries

arvindcheru:
Proposed Changes:
- SWDEV-310152: 
- Single version package: add ldconfig, no RPATH/RUNPATH in either binaries or libraries
- Multi version package: use RPATH for binaries, no RPATH/RUNPATH for libraries
- amdclang/clang/hipcc compilers shall not add RPATH/RUNPATH to produced binaries or libraries unless explicitly requested by an option
- Versioning scripts will take care of adding rpath to binaries for multi version package

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -586,8 +586,8 @@ fi
   if [[ "${build_relocatable}" == true ]]; then
     CXX=${compiler} ${cmake_executable} ${cmake_common_options[@]} ${cmake_client_options[@]} -DCPACK_SET_DESTDIR=OFF -DCMAKE_INSTALL_PREFIX="${rocm_path}" \
     -DCMAKE_PREFIX_PATH="${rocm_path};${rocm_path}/hip;$(pwd)/../deps/deps-install;${cuda_path};${cmake_prefix_path}" \
-    -DCMAKE_SHARED_LINKER_FLAGS="${rocm_rpath}" \
-    -DCMAKE_EXE_LINKER_FLAGS=" -Wl,--enable-new-dtags -Wl,--rpath,${rocm_path}/lib:${rocm_path}/lib64" \
+    -DCMAKE_SHARED_LINKER_FLAGS="" \
+    -DCMAKE_EXE_LINKER_FLAGS=" -Wl,--disable-new-dtags -Wl,--rpath,${rocm_path}/lib:${rocm_path}/lib64" \
     -DROCM_DISABLE_LDCONFIG=ON \
     -DROCM_PATH="${rocm_path}" ../..
   else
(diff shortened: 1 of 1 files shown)


cgmb:
## Quick summary

These flags set RUNPATH:

    -Wl,--enable-new-dtags -Wl,--rpath,${ROCM_RPATH}`

These flags set RPATH:

    -Wl,--rpath,${ROCM_RPATH}

or

    -Wl,--disable-new-dtags -Wl,--rpath,${ROCM_RPATH}

### Difference between RPATH and RUNPATH

To quote the [Debian wiki](https://wiki.debian.org/RpathIssue):

 >The dynamic linker will look for a matching library in the following locations, in this order, which can be changed (see the footnotes below):
> 
> 1.    the DT_RPATH dynamic section attribute of the library causing the lookup
> 1.    the DT_RPATH dynamic section attribute of the executable
> 1.    the LD_LIBRARY_PATH environment variable, unless the executable is setuid/setgid.
> 1.    the DT_RUNPATH dynamic section attribute of the executable
> 1.    /etc/ld.so.cache
> 1.    base library directories (/lib and /usr/lib) 

More information can be found with [`man ld.so`](https://man7.org/linux/man-pages/man8/ld.so.8.html).

This will mean that `LD_LIBRARY_PATH` will no longer work, though you can still use `LD_PRELOAD`. As I understand it, the motivation is so that executables don't need to set RUNPATH themselves.

These RPATH / RUNPATH attributes are only set in hipsolver when you use `./install.sh -r`. Though note that [hipcc itself adds flags](https://github.com/ROCm-Developer-Tools/HIP/blob/5664a6299f75e770c876017f93b08f96d8e98ae8/bin/hipcc.pl#L723) that enable RUNPATH (in ROCm 5.3 and earlier) or RPATH (in unreleased versions on the develop branch) on all shared libraries, unconditionally.

These changes to hipSOLVER look fine, though I find the behaviour of hipcc to be confusing. It results in binaries that are built with hipcc having a different lookup pattern than when built with a standard C++ compiler. It's also going to be very weird with HIP is available as a system library and hipcc starts adding `-Wl,-rpath=/usr/lib` to the libraries it builds. There's a _lot_ of libraries besides ROCm in /usr/lib and that may have unexpected consequences.

arvindcheru:
> 

Thanks for the detailed comments.
There were important discussions around this change in progress, summary of them at present, we have below points to be taken care.
- no RPATH/RUNPATH in both binaries and libraries
- Single version package add ldconfig
- multi version package use patchelf/chrpath to add rpath for binaries
- HIP should not to add any rpath/runpath by default 
- llvm/clang not to add rpath/runpath by default

This PR put on hold for now, as above approach,  dependent changes/release plans/target are getting finalized. Will update on this.

cgmb:
Thanks for the update, @arvindcheru! That's great news. It completely addresses my concerns.

cgmb (approved):

qjojo (approved):

PR #138, opened 2022-11-15
Move FindHIP.cmake into the project

cgmb:
This change copies FindHIP.cmake into the project (which is the typical way a find module is used). To quote [the CMake documentation on module mode searching](https://cmake.org/cmake/help/v3.22/command/find_package.html#search-modes):

> The `Find<PackageName>.cmake` file is not typically provided by the package itself. Rather, it is normally provided by something external to the package, such as the operating system, CMake itself, or even the project from which the find_package() command was called.

I then cut out most of the deprecated features of the FindHIP.cmake module (which we don't use anyway) and adapted it to be more similar to hip-config.cmake. My goal is to require as few changes as possible to our library build systems (or for users of our library build systems) when hip-config.cmake gets NVIDIA platform support.

The Findhip.cmake module acts as a wrapper around hip-config.cmake and as a replacement for  [FindHIP.cmake](https://github.com/ROCm-Developer-Tools/HIP/blob/rocm-5.3.2/cmake/FindHIP.cmake). Once hip-config.cmake is good enough that it doesn't need a wrapper or replacement, we can delete Findhip.cmake entirely and there should be few or no other changes required.


library/src/CMakeLists.txt
```diff
@@ -110,23 +110,20 @@ if( NOT USE_CUDA )
...
-
-  target_link_libraries( hipsolver PRIVATE ${CUDA_cusolver_LIBRARY} )
-
-  # External header includes included as system files
-  target_include_directories( hipsolver
-    SYSTEM PRIVATE
-      $<BUILD_INTERFACE:${HIP_INCLUDE_DIRS}>
-      $<BUILD_INTERFACE:${CUDA_INCLUDE_DIRS}>
-  )
+  target_link_libraries( hipsolver PRIVATE hip::host ${CUDA_cusolver_LIBRARY} )
```
cgmb, line 126: `hip::host` should be `PUBLIC` but HIP needs to export the target. We need SWDEV-243403 to be able to do that.

cgmb:
This is more or less the direction that the libraries are going in. The HIP Runtime has added NVIDIA platform support to hip-config.cmake and now provides the HIP_PLATFORM variable, so we won't even need to do it ourselves. However, this PR is very out-of-date in terms of conflicts, so I will be closing it. We'll create a new PR to begin using the improvements made by the HIP Runtime.

PR #141, opened 2022-12-09
Added gesvdj to the regular API

tfalders:
This PR adds gesvdj to the regular (and Fortran) API, and updates gesvdj to call the rocSOLVER version of this function.


jzuniga-amd (approved):
LGTM!
Nice job. I realize now how convoluted was to make it work before when calling the classic gesvd in rocsolver :)

cgmb (approved):

tfalders:
CI is failing because the rocSOLVER artifact is out-of-date and doesn't have the necessary functions. Unfortunately, I don't think we have enough time to wait for one to be built, and I can't log in to Jenkins to start a new build anyway, so it seems that I may have to merge without any CI tests against this PR. Let me know if there are any objections.

PR #147, opened 2023-02-15
Improve error reporting for zero-tolerance test failures

cgmb:
The hipsolver tests report certain error conditions (like unexpected info values) by adding a positive integer to the measured error. This can make it difficult to understand what is wrong when reviewing the logs for a failed set of tests (especially when there are multiple properties that could all increment the error).

The Google Test EXPECT functions will check a condition, but allow the test to continue. This means that an error can be recorded at the place where it occurs, without losing data.

This matches the changes made to rocsolver in cac76478d3a2fb860.

### DIFF
--- a/clients/include/testing_gebrd.hpp
+++ b/clients/include/testing_gebrd.hpp
@@ -1,5 +1,5 @@
 /* ************************************************************************
- * Copyright (C) 2020-2022 Advanced Micro Devices, Inc. All rights reserved.
+ * Copyright (C) 2020-2023 Advanced Micro Devices, Inc. All rights reserved.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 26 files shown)


cgmb:
(I noticed this would be useful for SWDEV-383316, so I ported it over while we were talking this afternoon.)

cgmb:
[Ugh](https://github.com/ROCmSoftwarePlatform/hipSOLVER/pull/147/commits/419a6c2a523d3be585f255481fe129bfe99477f0). The precheckin formatter uses the wrong clang-format by default, and it undid my fix.

cgmb:
It seems that SYGVDX_COMPAT has issues on a number of architectures. I don't think this failure is caused by my changes.
```
[ RUN      ] checkin_lapack/SYGVDX_COMPAT.__double/31
../../../../clients/include/testing_sygvdx_hegvdx.hpp:771: Failure
Expected equality of these values:
  hNev[b][0]
    Which is: 4
  hNevRes[b][0]
    Which is: 5
where b = 0
../../../../clients/include/testing_sygvdx_hegvdx.hpp:1369: Failure
Expected: ((max_error)) <= ((2 * n)*get_epsilon<T>()), actual: 1 vs 2.22045e-14
```

tfalders (approved):
LGTM!

jzuniga-amd (approved):

PR #148, opened 2023-03-02
Added hipsolverRf API

tfalders:
This PR adds hipsolverRf as a new compatibility API to hipSOLVER.

At the request of @cgmb, I'm opening as a draft PR for an early look.


clients/gtest/csrrf_refactlu_gtest.cpp
```diff
@@ -0,0 +1,107 @@
...
+
+using ::testing::Combine;
+using ::testing::TestWithParam;
+using ::testing::Values;
+using ::testing::ValuesIn;
+using namespace std;
+
+typedef std::tuple<int, int> csrrf_refactlu_tuple;
+
+// case when n = 0 and nnz = 0 also execute the bad arguments test
```
jzuniga-amd, line 15: there is no nnz=0 case value

clients/gtest/csrrf_refactlu_gtest.cpp
```diff
@@ -0,0 +1,107 @@
...
+    virtual void SetUp() {}
+    virtual void TearDown() {}
+
+    template <typename T>
+    void run_tests()
+    {
+        Arguments arg = csrrf_refactlu_setup_arguments(GetParam());
+
+        if(arg.peek<rocblas_int>("n") == 0 && arg.peek<rocblas_int>("nnzA") == 0)
+            testing_csrrf_refactlu_bad_arg<T>();
```
jzuniga-amd, line 72: it should be nnzA == 60 or something else

clients/include/testing_csrrf_refactlu.hpp
```diff
@@ -0,0 +1,604 @@
...
+                                 int*                indT,
+                                 T                   valT,
+                                 int*                pivP,
+                                 int*                pivQ)
+{
+    // N/A
+}
+
+template <typename T>
+void testing_csrrf_refactlu_bad_arg()
```
jzuniga-amd, line 46: all this code is unnecessary,  right?
I think you just need to test that the wrapper returns error when the handle is null

clients/include/testing_csrrf_refactlu.hpp
```diff
@@ -0,0 +1,604 @@
...
+                                Uh&                 hpivP,
+                                Uh&                 hpivQ,
+                                double*             gpu_time_used,
+                                double*             cpu_time_used,
+                                const int           hot_calls,
+                                const bool          perf,
+                                const fs::path      testcase)
+{
+    *cpu_time_used = nan(""); // no timing on cpu-lapack execution
+    *gpu_time_used = nan(""); // no timing on gpu-lapack execution
```
jzuniga-amd, line 353: we could be doing the setup, analysis and resetvalues only once, and timing the actual call to refRefactor. Not a big deal, something to think about.

clients/include/testing_csrrf_solve.hpp
```diff
@@ -0,0 +1,601 @@
...
+                              const int           nnzT,
+                              int*                ptrT,
+                              int*                indT,
+                              T                   valT,
+                              int*                pivP,
+                              int*                pivQ,
+                              T                   B,
+                              const int           ldb)
+{
+    // N/A
```
jzuniga-amd, line 22: Same here, you may want to test when the handle, B, pivP or pivQ are null, etc.

clients/include/testing_csrrf_solve.hpp
```diff
@@ -0,0 +1,601 @@
...
+                             Th&                 hB,
+                             Th&                 hX,
+                             double*             gpu_time_used,
+                             double*             cpu_time_used,
+                             const int           hot_calls,
+                             const bool          perf,
+                             const fs::path      testcase)
+{
+    *cpu_time_used = nan(""); // no timing on cpu-lapack execution
+    *gpu_time_used = nan(""); // no timing on gpu-lapack execution
```
jzuniga-amd, line 317: The same here: if we want, we can be timing the actual call to rfSolve

jzuniga-amd (commented):
LGTM. Good job!!
Just a couple of comments here...
I also have a PR to your branch with a few minor modifs to the documentation. Please have a look.

clients/CMakeLists.txt
```diff
@@ -47,27 +47,69 @@ list( APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake )
 
 include( build-options )
 
+# Linking lapack library requires fortran flags
+set( THREADS_PREFER_PTHREAD_FLAG ON )
+find_package( Threads REQUIRED )
+find_package( cblas REQUIRED CONFIG )
+
+if( NOT TARGET hipsolver )
+  find_package( hipsolver REQUIRED CONFIG PATHS ${ROCM_PATH}/hipsolver /opt/rocm/hipsolver )
```
cgmb, line 56: These paths are outdated. They should be `${ROCM_PATH} /opt/rocm`.

jzuniga-amd (approved):

qjojo (approved):

library/src/amd_detail/hipsolver_refactor.cpp
```diff
@@ -0,0 +1,952 @@
...
+    double *     dValU, *hValU;
+
+    rocblas_int *dPtrLU, *hPtrLU;
+    rocblas_int *dIndLU, *hIndLU;
+    double *     dValLU, *hValLU;
+
+    rocblas_int* dP;
+    rocblas_int* dQ;
+
+    uint8_t *d_buffer, *h_buffer;
```
cgmb, line 82: A buffer should be `void*` (or `char*` / `unsigned char*`). In practice, `uint8_t` is pretty much always a typedef for `unsigned char` and when that is the case, it inherits all the special exceptions to the aliasing rules that `char` gets, but that is not guaranteed.

library/src/amd_detail/hipsolver_refactor.cpp
```diff
@@ -0,0 +1,952 @@
...
+        , batch_count(0)
+        , effective_zero(0.0)
+        , boost_val(0.0)
+        , d_buffer(nullptr)
+        , h_buffer(nullptr)
+    {
+    }
+
+    // Allocate device memory
+    hipsolverStatus_t malloc_device(int n, int nnzA, int nnzL, int nnzU)
```
cgmb, line 106: This is a lot more gnarly than rocsolver. I don't see any obvious issues with this, but I definitely was wrong when I assumed that the resource management in hipSOLVER was similar to rocSOLVER.

library/include/internal/hipsolver-refactor.h
```diff
@@ -0,0 +1,215 @@
+/* ************************************************************************
+ * Copyright (C) 2023 Advanced Micro Devices, Inc.
+ * ************************************************************************ */
+
+#ifndef HIPSOLVER_REFACTOR_H
+#define HIPSOLVER_REFACTOR_H
+
+#include "hipsolver-types.h"
+
+typedef void* hipsolverRfHandle_t;
```
cgmb, line 10: We could maybe make this an opaque struct pointer, but this is consistent with our current handle. I think we want to _consider_ making this type-safe in ROCm 6.0, but I actually have a bunch of open questions about ABI compatibility between the rocm and cuda variants of HIP and the HIP math libraries that I want to get answered before making a decision there.

cgmb (approved):
I have a number of comments here and there, and it would be nice to copy across the whitespace cleanup on the test data from rocsolver. However, there's nothing that must be addressed. LGTM.

library/src/amd_detail/hipsolver_refactor.cpp
tfalders: I seem to remember hearing once that I couldn't trust `char` to be 8 bits in length on all compilers/architectures, so I thought this might be safer since it's explicit. I'm happy to change it if you believe it's safe.

library/include/internal/hipsolver-refactor.h
tfalders: Yeah, my plan was to keep things consistent with the existing `typedefs` for now, and then transition to opaque pointers in the next major version release.

library/src/amd_detail/hipsolver_refactor.cpp
cgmb: That's a good thought, but it doesn't actually matter. These buffers are allocated by malloc / hipMalloc, so the size you request is denominated in bytes. The type pointed to by `{d,h}_buffer` is never used during the size calculation.

---
It's a bit of a tangent, but I might as well explain why `uint8_t` would have the opposite effect of what you're expecting.

If you used `new uint8_t[]` rather than malloc, then the size of the array type would be part of the calculation of the allocation size, but that would actually make your size calculation incorrect. On some esoteric platform where `CHAR_BIT == 4`, then `sizeof(uint8_t) == 2`. The sizes of the elements are calculated with `sizeof(T) * count`, so you'd end up allocating double the required memory by creating an array of `uint8_5[sizeof(T) * count]`. And, you'd have the opposite problem on a platform where `CHAR_BIT > 8`... but then `uint8_t` wouldn't exist on that platform so the program would not compile anyway.

This is a long way to say: you never need to worry about `CHAR_BIT` when doing these sorts of allocation size calculations. `char` is one byte by definition, and all calculations are done in bytes so it doesn't matter how big a byte is.

PR #172, opened 2023-06-13
Convert void pointer typedefs into opaque types

tfalders:
This PR converts hipsolverHandle_t, hipsolverRfHandle_t, hipsolverGesvdjInfo_t, and hipsolverSyevjInfo_t from void pointers into pointers to opaque types. This should improve type safety from the user perspective.

Unfortunately, I've not been able to get this to work in a way that is non-invasive. In particular, I've had to add argument checks to all functions to make sure these opaque pointers are not null before being dereferenced. However, this does have the added benefit of allowing us to enable some of the bad_arg tests with the cuSOLVER backend and improving robustness to null pointers. At some point, it might be worth considering adding null pointer checks for all pointers sent to the cuSOLVER backend.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ Full documentation for hipSOLVER is available at [hipsolver.readthedocs.io](http
 ### Added
 ### Optimized
 ### Changed
+- hipsolverHandle_t, hipsolverRfHandle_t, hipsolverGesvdjInfo_t, and hipsolverSyevjInfo_t are now opaque types instead of void pointers.
+
 ### Deprecated
 ### Removed
 ### Fixed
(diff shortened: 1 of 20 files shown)


cgmb:
> Unfortunately, I've not been able to get this to work in a way that is non-invasive.

Ah. So previously we could just directly hand out a `static_cast<void*>(_rocblas_handle*)` or `static_cast<void*>(cusolverDnContext*)`. However, we can't have `struct hipsolverHandle { _rocblas_handle handle };`, so if we give out a `hipsolverHandle*`, we have to add another level of indirection.

tfalders:
Superseded by https://github.com/ROCmSoftwarePlatform/hipSOLVER/pull/177

PR #178, opened 2023-06-24
Set rocsolver minimum version

cgmb:
Search for rocSOLVER first so that it can impose version requirements on rocBLAS, if needed.


cgmb:
As shown by https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/586, there are some tricky issues to consider when checking the version. Let's shelve this for now.

PR #180, opened 2023-07-18
Use hipBLAS enums in hipSOLVER

tfalders:
This PR should make it easier to hipify cuSOLVER code by changing hipSOLVER to use hipBLAS enums in the same way cuSOLVER uses cuBLAS enums. This will require a companion PR to be opened on hipBLAS to prevent name conflicts.

My initial tests when using hipSOLVER and hipBLAS together had some errors, but on @doctorcolinsmith's suggestion I purged the build directory and built from scratch, and that seems to have resolved the errors. I will probably conduct further tests just to be safe.

EDIT: Since this is an API change, I've bumped the major version number.


CMakeLists.txt
```diff
@@ -89,7 +89,7 @@ include( ROCMInstallSymlinks )
 include( ROCMClients )
 include( ROCMHeaderWrapper )
 
-set ( VERSION_STRING "1.9.0" )
+set ( VERSION_STRING "2.0.0" )
```
jzuniga-amd, line 92: why the major version update?

tfalders: I think any code written with just hipSOLVER shouldn't be affected by the change, but it's an API change that will affect any code written with both hipSOLVER and hipBLAS. We're breaking compatibility with previous versions of hipBLAS, so I thought that justified a major version update.

tfalders: Also, I believe hipBLAS is updating their major version number as well. This'll make it easier to tell users which versions of hipSOLVER and hipBLAS work together, i.e. hipSOLVER v2 goes together with hipBLAS v2, and doesn't work with hipBLAS v1.

clients/gtest/CMakeLists.txt
```diff
@@ -87,6 +96,10 @@ target_link_libraries( hipsolver-test PRIVATE
   Threads::Threads
   roc::hipsolver
 )
+if( hipblas_FOUND )
+  target_link_libraries( hipsolver-test PRIVATE roc::hipblas )
+  target_sources( hipsolver-test PRIVATE hipblas_include1_gtest.cpp hipblas_include2_gtest.cpp )
+endif( )
```
cgmb, line 102: Rather than checking `hipblas_FOUND`, could we add an option that the user needs to explicitly set? We can set it to `OFF` by default to minimize dependencies required for third-parties to build and test hipSOLVER, but pass the flag to enable it in the .jenkins groovy files (much like is done with `WERROR` on the rocSOLVER CI). This will prevent it from being enabled by mistake just because hipBLAS happened to be available on the build machine.

library/include/internal/hipsolver-types.h
```diff
@@ -33,25 +33,71 @@ typedef enum
     HIPSOLVER_STATUS_ZERO_PIVOT        = 12,
 } hipsolverStatus_t;
 
-// set the values of enum constants to be the same as those used in cblas
+#ifndef HIPBLAS_OPERATION_DECLARED
+#define HIPBLAS_OPERATION_DECLARED
```
cgmb, line 37: We could do more to make this compatible with older versions of hipBLAS. There are a number of specific cases we could make compatible, although I don't think there's anything that would make it compatible in every case.

For example, when compiling with C++17 or a compiler that supports the extension, we could use `__has_include` and include `<hipblas/hipblas.h>` if available instead of using forward-declarations. Or, we could skip declaring the enums if `hipblasVersionMajor` is defined (since this indicates that the hipblas header has already been included). Those would all extend the source-compatibility of hipSOLVER with older versions of hipBLAS in a variety of cases. 

However, that would certainly be more complex and it could itself introduce more problems for users if we made any mistakes, so I'm fine with leaving this as-is.

cgmb (commented):
I'm skeptical of the value of the tests added for this. I know that you had some issues during the development of this feature that were fixed once you did a clean build, but have ever seen _these tests_ fail after you wrote them?

I'm fine with including them for now, but maybe let's limit them to the math-ci by default and review their costs vs. benefits in a few releases after we have some data.

clients/gtest/CMakeLists.txt
cgmb: Just wanted to highlight this comment, since I think it is the most important comment. I probably should have focused my comments on this topic, but brevity is not my strength.

tfalders: Good idea. I was thinking of adding a build option for my upcoming sparse PR as well, so this change should make things more consistent.

cgmb (approved):

qjojo (approved):

PR #184, opened 2023-08-17
Fix info tests for unitary generators

cgmb:
The operator++ has greater precedence than operator*, so the original form of this check merely incremented the pointer and then dereferenced it. On Windows, that resulted in the warning:

```cpp
<builddir>/clients/benchmarks/../include/testing_ormtr_unmtr.hpp:390:9: warning: expression result unused [-Wunused-value]
        *max_err++;
        ^~~~~~~~~~
```


tfalders (approved):
Nice catch! Thanks

jzuniga-amd (approved):

PR #185, opened 2023-08-17
Added csrlsvchol

tfalders:
This PR adds hipSOLVER wrappers for csrlsvchol and csrlsvcholHost. By default, the library now has dependencies on rocSPARSE and SuiteSparse, and hipsolver-test has a dependency on hipSPARSE. These can be disabled with the CMake option `BUILD_WITH_SPARSE` or the install script's `--no-sparse` flag.

The only thing missing right now is the GPU solve. Since rocSPARSE doesn't have a fully functional Cholesky factorization yet, I'm using SuiteSparse's Cholesky factorization even in the GPU method; which requires reading the data from the GPU and then sending it back after the factorization. Additionally, SuiteSparse doesn't have support for single-precision arrays, so I'm having to copy the original data to and from double-precision arrays.


CHANGELOG.md
```diff
@@ -7,8 +7,8 @@ Full documentation for hipSOLVER is available at [hipsolver.readthedocs.io](http
 - Added compatibility API with hipsolverSp prefix
 - Added compatibility-only functions
   - csrlsvchol
-    - hipsolverSpScsrlsvcholHost, hipsolverSpDcsrlsvcholHost, hipsolverSpCcsrlsvcholHost, hipsolverSpZcsrlsvcholHost
```
EdDAzevedo: The documentation for CHOLMOD mentions it has support for two different storage formats for complex matrices, one is the traditional interleave of real and imag parts, another format separately stores all real part in one matrix, and all imaginary part in another matrix.

clients/include/utility.hpp
```diff
@@ -527,6 +527,32 @@ void print_matrix(T* CPU_result, T* GPU_result, int m, int n, int lda)
                    double(GPU_result[i + j * lda].imag()));
 }
 
+/*! \brief  Debugging purpose, print out sparse matrix, not valid in complex number  */
+template <typename T, std::enable_if_t<!is_complex<T>, int> = 0>
+void print_sparse_matrix(int m, int n, int* csrRowPtr, int* csrColInd, T* csrVal)
```
EdDAzevedo, line 532: There is a also cholmod_print_sparse() routine to print a sparse matrix, but perhaps in compressed sparse column format.   Just FYI.

tfalders: I attempted to use that method, but the only thing it was printing out were the matrix dimensions and a validity check. If there's a way to get it to print the matrix elements, I haven't found how.

CHANGELOG.md
tfalders: I'm leaving out complex precision support from the current release because https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/582 is also leaving out complex support. I'm hoping that it won't be too difficult to get the complex methods working using SuiteSparse once it comes time to add them in.

qjojo (approved):

library/src/amd_detail/hipsolver_sparse.cpp
```diff
@@ -0,0 +1,909 @@
...
+
+    if((status = rocsolver_create_rfinfo(&sp->rfinfo, sp->handle)) != rocblas_status_success)
+    {
+        rocblas_destroy_handle(sp->handle);
+        delete sp;
+        return rocblas2hip_status(status);
+    }
+
+    int ret;
+    if((ret = cholmod_start(&sp->c_handle)) != true)
```
cgmb, line 234: I'm not 100% certain of how the promotions are done for this, but I would avoid comparing int and bool. Also, the `ret` variable isn't used.

.jenkins/debug.groovy
```diff
@@ -18,7 +18,7 @@ def runCI =
 
     // customize for project
     prj.paths.build_command = buildCommand
-    prj.libraryDependencies = ['rocBLAS-internal', 'rocSPARSE-internal', 'rocSOLVER', 'hipBLAS']
+    prj.libraryDependencies = ['rocBLAS', 'rocSPARSE', 'rocSOLVER', 'hipBLAS', 'hipSPARSE']
```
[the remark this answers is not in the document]
tfalders: That's a good question. I don't remember why I changed it. I'll change it back.

library/src/CMakeLists.txt
```diff
@@ -120,6 +124,28 @@ if( NOT USE_CUDA )
...
+      else( )
+        find_package( rocsparse REQUIRED CONFIG PATHS /opt/rocm /opt/rocm/rocsparse )
+      endif( )
+    endif( )
+
+    target_link_libraries( hipsolver PRIVATE roc::rocsparse suitesparseconfig cholmod )
+    target_include_directories( hipsolver
+      SYSTEM PRIVATE
+      $<BUILD_INTERFACE:/usr/include/suitesparse/>
+    )
```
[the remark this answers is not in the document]
cgmb: It seems that the most recent versions of suitesparse do have config.cmake, but we will probably need to write a find module for older versions.

cgmb (approved):

.jenkins/debug.groovy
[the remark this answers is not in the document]
tfalders: Ah, it looks like there was a reason why I changed it: https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/630

Should I restore the change, do you think?

jzuniga-amd (commented):
I opened PR with some minor modifs to the documentation. Please have a look.

docs/userguide/install.rst
```diff
@@ -12,11 +12,14 @@ Installation
...
 
-Download pre-built packages from `ROCm's package servers <https://docs.amd.com/bundle/ROCm-Installation-Guide-v5.4.3/page/Introduction_to_ROCm_Installation_Guide_for_Linux.html>`_. Release notes
-are available on the releases tab of the `library's github page <https://github.com/ROCmSoftwarePlatform/hipSOLVER>`_.
+Download pre-built packages from `ROCm's package servers <https://docs.amd.com/bundle/ROCm-Installation-Guide-v5.4.3/page/Introduction_to_ROCm_Installation_Guide_for_Linux.html>`_.
+Release notes are available on the releases tab of the `library's github page <https://github.com/ROCmSoftwarePlatform/hipSOLVER>`_.
 
 * `sudo apt update && sudo apt install hipsolver`
 
+Note that the pre-built packages depend on the third-party library SuiteSparse, which must be installed on the system prior to installing hipSOLVER.
+SuiteSparse can be installed using the package manager of most distros.
```
jzuniga-amd, line 21: Yeah, that's clearer now. 
Should we also add a comment like:
"For now, if a user does not need the sparse functionality and does not want to install suitsparse, it is recommended to build hipSOLVER from source" ?

jzuniga-amd (approved):
LGTM

tfalders:
CI is failing because of the unavailability of dependency artifacts, but in order to meet our deadlines I'm going to force the merge.

PR #193, opened 2023-09-20
Specify build.os in the .readthedocs.yaml config (replaces build.image)

samjwu:
https://docs.readthedocs.io/en/stable/config-file/v2.html

build.image (which is implicitly used if neither is specified) will be deprecated in October 2023

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -9,6 +9,10 @@ sphinx:
 formats: [htmlzip]
 
 python:
-   version: "3.8"
    install:
    - requirements: docs/.sphinx/requirements.txt
+
+build:
+   os: ubuntu-20.04
+   tools:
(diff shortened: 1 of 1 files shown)


.readthedocs.yaml
```diff
@@ -9,6 +9,10 @@ sphinx:
 formats: [htmlzip]
 
 python:
-   version: "3.8"
    install:
    - requirements: docs/.sphinx/requirements.txt
+
+build:
+   os: ubuntu-20.04
```
cgmb, line 16: Is there any reason to prefer 20.04 to 22.04? I'd be tempted to specify the newer one, as it will probably be supported longer.

samjwu: changed to 22.04

tfalders (approved):

PR #195, opened 2023-09-26
Use consistent test data

tfalders:
hipSOLVER's test data changes from run to run, occasionally causing intermittent test failures on CI. This is particularly noticeable in SYGVDX, where some instances of the test data have eigenvalues near the cutoff point, causing rocSOLVER and vanilla LAPACK to disagree on the number of eigenvalues in the allowed range.

This PR changes the test data initialization to be more like rocSOLVER, and to be consistent between test runs.

### DIFF
--- a/clients/include/utility.hpp
+++ b/clients/include/utility.hpp
@@ -390,8 +390,7 @@ public:
 template <typename T>
 T random_generator()
 {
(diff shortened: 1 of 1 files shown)


jzuniga-amd (approved):

cgmb (approved):
I'm always in favour of defaulting to reproducibility.

PR #197, opened 2023-09-27
Update changelog and version for ROCm 5.7.1

cgmb:
These changes are necessary to ensure that each release of the hipSOLVER library with a unique commit hash (1) has an entry in the changelog and (2) has a distinct version number.

I should have requested that these changes be included in #192. Mea culpa.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 Full documentation for hipSOLVER is available at [hipsolver.readthedocs.io](https://hipsolver.readthedocs.io/en/latest/).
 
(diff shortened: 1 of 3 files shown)


cgmb:
You know what, I can probably be more specific in the changelog. "Exclude hipsolver include folder from ASAN package" was the title of the PR, but the problem was specifically with `hipsolver_module.f90` (which is why this wasn't necessary for rocsolver). I'll update that.

cgmb:
Much better. Please review and let me know if you have any concerns.

tfalders (approved):

jzuniga-amd (approved):

PR #200, opened 2023-10-19
SWDEV-421983: Fix the computation of U and V matrices in gesvda

xinyazhang:
Fixes SWDEV-421983

Per specification of [cusolverDn<t>gesvdaStridedBatched](https://docs.nvidia.com/cuda/cusolver/index.html?highlight=gesvda#cusolverdn-t-gesvdastridedbatched):
> jobz = `CUSOLVER_EIG_MODE_NOVECTOR` : Compute singular values only; jobz = `CUSOLVER_EIG_MODE_VECTOR` : Compute singular values and singular vectors.

and
[rocblas_svect](https://rocmdocs.amd.com/projects/rocSOLVER/en/latest/api/types.html#rocblas-svect)
> * enumerator rocblas_svect_all
>       + The entire associated orthogonal/unitary matrix is computed.
> * enumerator rocblas_svect_singular
>       + Only the singular vectors are computed and stored in output array.
> * enumerator rocblas_svect_overwrite
>       + Only the singular vectors are computed and overwrite the input matrix.
> * enumerator rocblas_svect_none
>       +  No singular vectors are computed

This means 
* jobz = `CUSOLVER_EIG_MODE_NOVECTOR`, `hip2rocblas_evect2svect(jobz, xxx)` should return `rocblas_svect_singular`
* jobz = `CUSOLVER_EIG_MODE_VECTOR`, `hip2rocblas_evect2svect(jobz, xxx)` should return `rocblas_svect_all`
    + Consequently the `xxx` above should be the constant `0`

Additionally, `hip2rocblas_evect2svect(jobz, econ)` should return `rocblas_svect_none` when `econ=1`, because the in `_bufferSize` functions no pointer is provided and no computation should be performed.

### DIFF
--- a/library/src/amd_detail/hipsolver.cpp
+++ b/library/src/amd_detail/hipsolver.cpp
@@ -6194,8 +6194,8 @@ try
     // perform computation
     CHECK_ROCBLAS_ERROR(
         rocsolver_cgesvdx_strided_batched((rocblas_handle)handle,
-                                          hip2rocblas_evect2svect(jobz, 1),
-                                          hip2rocblas_evect2svect(jobz, 1),
+                                          hip2rocblas_evect2svect(jobz, 0),
+                                          hip2rocblas_evect2svect(jobz, 0),
                                           rocblas_srange_index,
                                           m,
                                           n,
--- a/library/src/amd_detail/hipsolver_conversions.cpp
+++ b/library/src/amd_detail/hipsolver_conversions.cpp
@@ -213,10 +213,10 @@ rocblas_svect_ hip2rocblas_evect2svect(hipsolverEigMode_t eig, int econ)
     switch(eig)
     {
     case HIPSOLVER_EIG_MODE_NOVECTOR:
-        return rocblas_svect_none;
+        return rocblas_svect_singular;
     case HIPSOLVER_EIG_MODE_VECTOR:
         if(econ)
-            return rocblas_svect_singular;
+            return rocblas_svect_none;
         else
             return rocblas_svect_all;
     default:


tfalders:
> * jobz = `CUSOLVER_EIG_MODE_NOVECTOR`, `hip2rocblas_evect2svect(jobz, xxx)` should return `rocblas_svect_singular`

`rocblas_svect_none` will compute the singular values, and no singular vectors will be computed. This matches the stated behaviour of  `CUSOLVER_EIG_MODE_NOVECTOR`.

> * jobz = `CUSOLVER_EIG_MODE_VECTOR`, `hip2rocblas_evect2svect(jobz, xxx)` should return `rocblas_svect_all`

`rocblas_svect_all` will compute the full m*m matrix U and n*n matrix V. According to the cuSOLVER documentation, U is an m*n matrix (with m>n), which means that only the singular vectors (not the full matrix) are being computed. This behaviour is provided by `rocblas_svect_singular`.

> Additionally, `hip2rocblas_evect2svect(jobz, econ)` should return `rocblas_svect_none` when `econ=1`, because the in `_bufferSize` functions no pointer is provided and no computation should be performed.

The calls to `rocblas_start_device_memory_size_query` ensure that the `_bufferSize` functions do not perform any computations. `rocblas_svect_none` does not prevent computations from being performed - as stated above, it indicates that all of the singular values, and none of the singular vectors, are to be computed.


It should be noted that gesvda with the rocSOLVER backend requires larger arrays than cuSOLVER does (see https://hipsolver.readthedocs.io/en/latest/userguide/usage.html#different-minimum-array-lengths). It's possible that this is the cause of [SWDEV-421983](http://ontrack-internal.amd.com/browse/SWDEV-421983).

xinyazhang:
Invalid, the problem is at rocSOLVER

PR #201, opened 2023-10-25
Standardize documentation

samjwu:
Relates to https://github.com/RadeonOpenCompute/rocm-docs-core/issues/330

Apply the following to repo docs:
add version number to documentation left navigation bar and page title
add an "About" section with a license page
enable htmlzip, pdf, epub formats when publishing on Read the Docs
set pdf title, author, copyright, and version
rename .sphinx/.doxygen to sphinx/doxygen
remove docBin from URL
update rocm-docs-core dependency
update dependabot config


docs/conf.py
```diff
@@ -4,14 +4,33 @@
...
+
+external_toc_path = "./sphinx/_toc.yml"
 
-docs_core = ROCmDocs("hipSOLVER Documentation")
-docs_core.run_doxygen()
+docs_core = ROCmDocs(left_nav_title)
+docs_core.run_doxygen(doxygen_root="doxygen", doxygen_path="doxygen/xml")
+docs_core.enable_api_reference()
 docs_core.setup()
+docs_core.setup()
```
tfalders, line 31: It looks like the setup call is duplicated? Also, I didn't see `enable_api_reference` in https://github.com/ROCmSoftwarePlatform/rocSOLVER/pull/626. Should it be present there as well?

tfalders (approved):

docs/conf.py
samjwu: Thanks for catching that. Since hipSOLVER documentation does not use [doxysphinx](https://pypi.org/project/doxysphinx/) these 2 extra lines are not needed.

```
docs_core.enable_api_reference()
docs_core.setup()
```

jzuniga-amd (approved):

PR #206, opened 2023-11-15
Temporarily disable hipBLAS tests on CI

tfalders:


cgmb (approved):
We have the same failures as hipBLAS on the NVIDIA platform CI, but it's still an improvement to have the AMD platform CI passing.

jzuniga-amd (approved):

PR #216, opened 2023-12-22
Update ReadtheDocs links to new documentation sites

samjwu:
Replaces ReadtheDocs for Community links (.io) (being retired) with for Business (.com)

Also applies some automatic Markdown formatting changes to the changelog

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,21 +1,29 @@
 # Change Log for hipSOLVER
 
-Full documentation for hipSOLVER is available at [hipsolver.readthedocs.io](https://hipsolver.readthedocs.io/en/latest/).
-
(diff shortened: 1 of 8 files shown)


CHANGELOG.md
tfalders, line 1: I'm not a fan of the changelog changes. It makes it much harder to isolate specific releases when viewing the raw file. Is there a specific rationale for the changes or can we revert?

tfalders (changes requested):
Looks good, though I have concerns about the changes to the changelog.

CHANGELOG.md
samjwu: It was mainly to address markdown linting violations (eg: they require a newline between each header and paragraph), but this can also be reverted.

samjwu: Reverted in latest. Only the doc link was updated.

tfalders (approved):

PR #226, opened 2024-02-07
cmake: add blas/solver/sparse as package deps if building static lib

evetsso:
When a static library is built, no linking is done.  (Basically static libraries are just archives of object files.)

So if `libfoo.a` depends on `libbar.a`, CMake just has to remember that when you link `libfoo.a`, it also needs to pull in `libbar.a`.  The problem is that CMake only knows about `libbar.a` if a `find_package(bar)` is done.  This PR fixes our generated `foo-config.cmake` file to find this dependency, when we're building a static library.

### DIFF
--- a/library/src/CMakeLists.txt
+++ b/library/src/CMakeLists.txt
@@ -96,6 +96,8 @@ if( WIN32 )
   endif( )
 endif( )
 
+set( static_depends )
+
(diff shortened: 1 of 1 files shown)


cgmb (approved):
LGTM.

tfalders (approved):

PR #232, opened 2024-02-23
Re-enable static library ci job

samjwu:
Requires static jobs for dependencies to pass

http://math-ci.amd.com/job/static-libraries/job/static/job/hipSOLVER/job/PR-232/

### DIFF
--- a/.jenkins/static.groovy
+++ b/.jenkins/static.groovy
@@ -15,7 +15,7 @@ def runCI =
 
     def prj  = new rocProject('hipSOLVER', 'StaticLibrary')
     prj.paths.build_command = './install.sh -cd --static -p /opt/rocm/lib/cmake'
-    prj.libraryDependencies = ['rocBLAS-internal', 'rocSPARSE-internal', 'rocSOLVER', 'hipSPARSE']
+    prj.libraryDependencies = ['rocBLAS', 'rocSPARSE', 'rocSOLVER', 'hipSPARSE']
     prj.defaults.ccache = true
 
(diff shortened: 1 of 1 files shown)


samjwu:
http://math-ci.amd.com/blue/organizations/jenkins/static-libraries%2Fstatic%2FhipSOLVER/detail/PR-232/2/pipeline

cgmb:
@jzuniga-amd, @tfalders, we should consider changing how the compatibility API is organized as it currently leads to very large binaries with static libraries. When linking static libraries, the smallest unit of logic that the linker will pull into a binary is an object file (i.e., the result of compiling a single translation unit / `.cpp` file). Since we have all the compatibility functions defined in a single translation unit, anyone using a single compatibility function ends up pulling in every compatibility function (which in turn pulls in every rocSOLVER function).

The definitions of the compatibility functions should be moved into the same translation units as the non-compatibility functions. (And, IMO, we should be using the [`alias` attribute](https://gcc.gnu.org/onlinedocs/gcc-4.1.2/gcc/Function-Attributes.html) for cases where the compatibility functions are identical to the non-compatibility functions in all but name.)

@samjwu, that of course, has nothing to do with your PR. It just reminded me.

samjwu:
http://math-ci.amd.com/blue/organizations/jenkins/static-libraries%2Fstatic%2FhipSOLVER/detail/PR-232/5/pipeline

passed on rhel8/ubuntu22

cgmb (approved):

PR #243, opened 2024-03-12
Added APIs to the hipsolver namespace

jibrealkhan:
- Added APIs to the hipsolver namespace.
- Changed the default namespace form std to hipsolver.


tfalders (approved):

jzuniga-amd (approved):

library/src/include/lib_macros.hpp
```diff
@@ -25,6 +25,17 @@
 
 #include "hipsolver.h"
 
+// Macro for Namespace
+#ifndef HIPSOLVER_BEGIN_NAMESPACE
+#define HIPSOLVER_BEGIN_NAMESPACE    \
+    namespace hipsolver {            \
+    inline namespace v2_2_0 {
```
tfalders: It occurs to me that we should either do the same thing as in rocSOLVER to keep the version number up-to-date, or add this to the bump_hipsolver_version script.

cgmb: It has been to use the version macros.

cgmb (approved):
I adjusted the inline namespace to use the hipsolver version macros, and I dropped the use of `using namespace hipsolver;`.

Also related to the static build, I notice that we have a _lot_ of functions in a handful of implementation files. That's going to cause a lot of bloat for the users of the static library. We should get an item on our backlog to break them up into smaller translation units, like in rocSOLVER.

cgmb:
Please review updated PR. If you strongly prefer the `using namespace hipsolver;`, please do speak up.

tfalders:
I think you might need to apply clang format

tfalders:
Hmmm... I do think I prefer `using namespace hipsolver`, but perhaps that's something we can think about when splitting up the large cpp files into separate units. Maybe it makes more sense to use `using namespace hipsolver` when it's only affecting a small compilation unit.

tfalders (approved):

jzuniga-amd (approved):

PR #246, opened 2024-03-22
Disable sparse build on Windows by default

cgmb:
*  Default BUILD_WITH_SPARSE to OFF on Windows
*  Cleanup include paths in hipsolver_sparse.cpp

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -110,7 +110,7 @@ option( BUILD_CODE_COVERAGE "Build hipSOLVER with code coverage enabled" OFF )
 option( BUILD_HIPBLAS_TESTS "Build additional tests to ensure hipBLAS and hipSOLVER are compatible (requires installed hipBLAS)" OFF )
 # BUILD_SHARED_LIBS is a cmake built-in; we make it an explicit option such that it shows in cmake-gui
 option( BUILD_SHARED_LIBS "Build hipSOLVER as a shared library" ON )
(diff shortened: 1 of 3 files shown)


EdDAzevedo (approved):
I am no expert in cmake configurations but the changes look reasonable to me.

clients/include/hipsolver_dispatcher.hpp
```diff
@@ -120,8 +122,10 @@ class hipsolver_dispatcher
             {"sygvdx", testing_sygvdx_hegvdx<API_COMPAT, false, false, T>},
             {"sygvj", testing_sygvj_hegvj<API_NORMAL, false, false, T>},
             {"sytrd", testing_sytrd_hetrd<API_NORMAL, false, false, T>},
+#ifdef HAVE_ROCSPRASE
```
tfalders: Spelling error on ROCSPARSE.

Have a look at https://github.com/ROCm/hipSOLVER/pull/242, since it contains my approach to the same issue. I think we should have a compile definition that is set in the benchmark client CMake, since IIRC it's possible to build the client without building the library.

cgmb: Good catch. Let's get your fix merged in.

tfalders (approved):

cgmb:
The Windows CI passed the build _and_ tests. :O

PR #252, opened 2024-04-08
Update suitesparse packages for Ubuntu 24.04

cgmb:
The suitesparse library has changed its ABI a few times since Ubuntu 22.04. The package names include the ABI major version, so they're a little different on the upcoming Ubuntu 24.04.

Ticket: SWDEV-455244

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -224,6 +224,8 @@ if( NOT USE_CUDA )
 
     if( SYSTEM_OS STREQUAL "centos" OR SYSTEM_OS STREQUAL "rhel" )
       list(APPEND hipsolver_pkgdeps "suitesparse")
+    elseif(SYSTEM_OS STREQUAL "ubuntu" AND SYSTEM_OS_VERSION VERSION_GREATER_EQUAL "24.04")
+      list(APPEND hipsolver_pkgdeps "libcholmod4" "libsuitesparseconfig7")
     else()
       list(APPEND hipsolver_pkgdeps "libcholmod3" "libsuitesparseconfig5")
     endif()


tfalders (approved):

jzuniga-amd (approved):

powderluv:
Can we take a dependency and build / link against our own version of suitesparse  so we don't hit variations with every distro update ?

cgmb:
If we used proper Debian package tooling, this information would be provided by the package tools automatically, rather than being something that we manually have to code logic for. Our CPack package logic is full of assumptions that may not necessarily be true if your build environment differs from the way that the AMD official builds are done.

In native tooling, the package name of the dependency is automatically handled by  `${shlibs:Depends}` [in the package Depends field of debian/control](https://salsa.debian.org/rocm-team/hipsolver/-/blob/debian/5.5.1-5/debian/control?ref_type=tags#L35).

CPack is inferior in a lot of ways. For example, you cannot use `apt-get build-dep rocsolver` with AMD's ROCm repos to install all of the packages needed to build rocSOLVER because CPack doesn't know that information. The manual tracking of minimum versions for the Depends field is also something that is only required with CPack and not with native tools. The detection of ABI breaking changes is also built into native tooling, but not in CPack... the list goes on.

I think we should just use better tools.

cgmb:
That isn't to say that CPack is awful or anything. It has been a useful tool and it's an easy way to support a variety of very different native packaging toolsets with some very simple code. My personal opinion is just that it's a good tool to make basic packages and we're starting to outgrow it.

cgmb:
More directly answering this question...
> Can we take a dependency and build / link against our own version of suitesparse so we don't hit variations with every distro update ?

For our static builds, we will probably have to do that for a number of dependencies. We could extend the system we set up for the static libraries to do the same for the dynamic libraries, but I think that's a question that needs a ROCm-wide architecture discussion.

cgmb:
The results are not getting reported to github, but the tests passed on CI.

haampie:
@cgmb I would appreciate a proper fix, cause being on ubuntu obviously does not mean you're using system ubuntu's suite sparse, it's completely unrelated to the distro what suite sparse a user might want to link to.

Can't you use `find_package` / `find_library`?

Otherwise we'll get an infinite number of PRs like https://github.com/spack/spack/pull/44473

haampie:
Ah, this may only be used for cpack. Nevermind.

But still, please use `find_library(...)` and create a target for libsuitesparse, instead of hoping `target_link_libraries(<target> <name>)` works, which links to some arbitrary libsuitesparse.

cgmb:
@haampie, the plan is to drop suitesparse at build-time and dlopen it instead, although the option to link against it at build-time will remain available. I'll update hipsolver to use `find_package` as that should work with recent versions of suitesparse, and fall back on `find_library`. My original intention was to do that for ROCm 6.1, but it's still on my todo list. If the dlopen changes continue to be deferred, I'll try to get those build improvements in separately.

> Ah, this may only be used for cpack. Nevermind.

Yeah, the CPack stuff unavoidably contains a lot of simplifying assumptions but won't affect third-party packagers.

PR #253, opened 2024-04-09
Update tests for new argument checking

tfalders:
Updates tests to match https://github.com/ROCm/rocSOLVER/pull/691


jzuniga-amd (approved):
Right, I forgot these should be changed here too. Thanks.

jmachado-amd (approved):

PR #255, opened 2024-04-12
Added 64-bit getrs

tfalders:
This PR adds the 64-bit method hipsolverDnXgetrs to the hipSOLVER compatibility API.


lpaoletti (approved):
The changes only impact one RST file. Changes look fine. 

Skimmed the other files. Seems good.

clients/gtest/getrs_gtest.cpp
```diff
@@ -114,22 +114,26 @@ class GETRS_BASE : public ::TestWithParam<getrs_tuple>
...
+class GETRS_FORTRAN : public GETRS_BASE<API_FORTRAN, int, int>
 {
 };
 
-class GETRS_COMPAT : public GETRS_BASE<API_COMPAT>
+class GETRS_COMPAT : public GETRS_BASE<API_COMPAT, int, int>
+{
+};
+
+class GETRS_COMPAT_64 : public GETRS_BASE<API_COMPAT, int64_t, size_t>
```
jmachado-amd, line 136: Just a comment on the asymmetry, `SIZE` is `int` when 32-bit, but `size_t` when 64-bit.

clients/rocblascommon/device_strided_batch_vector.hpp
```diff
@@ -189,6 +189,30 @@ class device_strided_batch_vector : public d_vector<T, PAD, U>
...
+    //! @param batch_index The batch index.
+    //! @return A non-mutable mutable pointer to the batch_index'th vector.
+    //!
+    const T* operator[](int64_t batch_index) const
+    {
+        return (this->m_stride >= 0)
+                   ? this->m_data + batch_index * this->m_stride
+                   : this->m_data + (batch_index + 1 - this->m_batch_count) * this->m_stride;
+    }
+
```
jmachado-amd: At some point in time `m_n`, `m_inc`, and `m_batch_count` will probably need to become `int64_t` as well.

clients/rocblascommon/host_strided_batch_vector.hpp
```diff
@@ -191,6 +191,30 @@ class host_strided_batch_vector
...
+    //! @brief Returns non-mutable pointer.
+    //! @param batch_index The batch index.
+    //! @return A non-mutable mutable pointer to the batch_index'th vector.
+    //!
+    const T* operator[](int64_t batch_index) const
+    {
+        return (this->m_stride >= 0)
+                   ? this->m_data + this->m_stride * batch_index
+                   : this->m_data + (batch_index + 1 - this->m_batch_count) * this->m_stride;
+    }
```
jmachado-amd: Similarly to the case of `device_strided_batch_vector.hpp`, at some point the `rocblas_int`s will probably have to become `int64_t`.

clients/include/utility.hpp
```diff
@@ -291,6 +291,43 @@ class hipsolver_local_syevj_info
...
+        hipsolverDnDestroyParams(m_info);
+    }
+
+    hipsolver_local_params(const hipsolver_local_params&) = delete;
+
+    hipsolver_local_params(hipsolver_local_params&&) = delete;
+
+    hipsolver_local_params& operator=(const hipsolver_local_params&) = delete;
+
+    hipsolver_local_params& operator=(hipsolver_local_params&&) = delete;
```
jmachado-amd, line 318: Just a question: any particular reason to `delete` the move operations? Asking because they would make sense semantically for this class.

jmachado-amd (approved):
The PR looks good to me! I see that the fix to `hipsolver-types.h` fixes the compilation, the other comments I have are minor. Approving because there is no need to implement them.

clients/gtest/getrs_gtest.cpp
tfalders: Yes, the 32-bit cuSOLVER APIs use `int` for the workspace size, while the 64-bit APIs use `size_t`.

clients/rocblascommon/device_strided_batch_vector.hpp
tfalders: Thanks for reminding me. I'll change over the container classes to be 64-bit as in rocSOLVER.

clients/include/utility.hpp
tfalders: I'm not terribly familiar with how move operations work, but I think the idea is to ensure that the class deconstructor can only ever be called once, since that will deallocate the wrapped Params object.

cgmb: The move operations could be implemented, but it's more complicated than just deleting them. Since we don't actually use the move operations, we might as well just keep it simple.

jzuniga-amd (approved):
LGTM

tfalders:
For some reason, CI status isn't being reported to github. It looks like all the tests have passed, so if there are no objections, I'll force the merge.

PR #259, opened 2024-04-16
Build system improvements

tfalders:
This PR brings hipSOLVER's build system more in line with hipBLAS's.

Although it would be nice to get it in for the next release, it's not essential. Don't feel pressured to review it for tomorrow.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,14 +39,14 @@ if( NOT DEFINED CMAKE_CONFIGURATION_TYPES AND NOT DEFINED CMAKE_BUILD_TYPE )
   set( CMAKE_BUILD_TYPE Release CACHE STRING "Choose the type of build, options are: None Debug Release RelWithDebInfo MinSizeRel." )
 endif()
 
-if ( NOT DEFINED CMAKE_Fortran_COMPILER AND NOT DEFINED ENV{FC} )
-  set( CMAKE_Fortran_COMPILER  "gfortran" )
-endif()
-
-project( hipsolver LANGUAGES CXX )
-if( UNIX )
-  enable_language( Fortran )
+if (NOT WIN32)
+    if ( NOT DEFINED CMAKE_Fortran_COMPILER AND NOT DEFINED ENV{FC} )
+      set( CMAKE_Fortran_COMPILER  "gfortran" )
+    endif()
+    set( fortran_language "Fortran" )
 endif( )
+
+project( hipsolver LANGUAGES CXX ${fortran_language} )
 
 # This finds the rocm-cmake project, and installs it if not found
 # rocm-cmake contains common cmake code for rocm projects to help setup and install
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -39,15 +39,15 @@ if( NOT DEFINED CMAKE_CONFIGURATION_TYPES AND NOT DEFINED CMAKE_BUILD_TYPE )
...
-  enable_language( Fortran )
+if (NOT WIN32)
+    if ( NOT DEFINED CMAKE_Fortran_COMPILER AND NOT DEFINED ENV{FC} )
+      set( CMAKE_Fortran_COMPILER  "gfortran" )
+    endif()
+    set( fortran_language "Fortran" )
 endif( )
 
+project( hipsolver LANGUAGES CXX ${fortran_language} )
+
```
cgmb, line 50: This seems more complicated than what it replaced.

CMakeLists.txt
```diff
@@ -112,24 +106,13 @@ option( BUILD_HIPBLAS_TESTS "Build additional tests to ensure hipBLAS and hipSOL
 option( BUILD_SHARED_LIBS "Build hipSOLVER as a shared library" ON )
 option( BUILD_WITH_SPARSE "Build hipSOLVER with sparse functionality and tests enabled (requires additional dependencies)" "${UNIX}" )
 option( BUILD_VERBOSE "Output additional build information" OFF )
-option( USE_CUDA "Look for CUDA and use that as a backend if found" OFF )
 option( HIPSOLVER_FIND_PACKAGE_LAPACK_CONFIG "Skip module mode search for LAPACK" ON )
+option( BUILD_FILE_REORG_BACKWARD_COMPATIBILITY "Build with file/folder reorg backward compatibility enabled" OFF )
```
cgmb, line 110: The approach used in hipBLAS for this is silly and it would be a step backward to adopt it. If we were to make any change, though, it would probably just be to remove the option entirely.

CMakeLists.txt
```diff
@@ -139,32 +122,59 @@ endif( )
...
+    message( DEPRECATION "USE_CUDA is deprecated (use environment variable HIP_PLATFORM=amd)" )
+  endif( )
+endif( )
 
 # Hip headers required of all clients; clients use hip to allocate device memory
-if( USE_CUDA)
+find_package( hip CONFIG PATHS ${ROCM_PATH} /opt/rocm )
+
+# support for cuda backend with hip < 6.0
+if( NOT hip_FOUND )
```
cgmb, line 140: The comment should really be hip < 5.7 because that's when the hip-nvidia-config.cmake implementation was added.

I'm not sure it's worth including this logic, since it's not going to work in a lot of situations (e.g., if you have both the AMD and NVIDIA tooling installed with ROCm 5.6, then the `find_package(hip)` will find the AMD platform tooling and you won't be able to build for the CUDA backend). If it was cost-free, then maybe that would be worth it anyway, but this fallback makes it more complicated to reason about the rest of the build.

If this is landing in ROCm 6.3, I don't think it's too unreasonable to require ROCm 5.7 or later to build for the CUDA backend at that point.

cgmb (changes requested):
The update to use config-mode search for the NVIDIA backend is an important build system change and we should keep it tightly focused so that it can be easily reverted. I know the inconsistent formatting drives you nuts, but we should keep those sorts of changes separate.

Sorry, Troy. I think we should talk about this one a bit more and push it to the next release cycle.

tfalders:
> The update to use config-mode search for the NVIDIA backend is an important build system change and we should keep it tightly focused so that it can be easily reverted. I know the inconsistent formatting drives you nuts, but we should keep those sorts of changes separate.

Yeah, I was going to focus on the build changes only, but then I got carried away after reformatting the pieces I borrowed from hipBLAS. Since all the changes (other than formatting) are based on something I saw in hipBLAS, any objections you have to changes in this PR suggest that we might want to propose tweaks to hipBLAS's build system to bring it more in line with hipSOLVER's.

PR #272, opened 2024-05-07
Merge release-staging/rocm-rel-6.2 into develop

cgmb:
This merges back the updates made on the ROCm 6.2 staging branch back into develop.


library/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@
 
 
 # This is incremented when the ABI to the library changes
-set( hipsolver_SOVERSION 0.1 )
+set( hipsolver_SOVERSION 0.2 )
```
tfalders: Can we update the SO version in this PR to 0.3?

cgmb: Sure. I also started escaping the regexes in `bump_hipsolver_version.sh`, as I noticed that `sed 's/0.4/0.5/g'` would change '(C) 2024' to '(C) 2025'.

EdDAzevedo: Perhaps you can try something like  "s/0\\.4/0\\.5/g"   so that "." means the "dot" and not be used to match any character?  Just a suggestion.

tfalders (approved):

jzuniga-amd (approved):

PR #286, opened 2024-06-07
Use find_library to search for cholmod

cgmb:
Use find_library to search for cholmod. Newer versions of suitesparse provide a CHOLMODConfig.cmake, so we can try searching for it first before adding these new find modules to the CMake module path and searching with `find_library` / `find_path`. The `CHOLMOD_ROOT` environment variable can be used as a hint both for config-based searches and find-module based searches.


cgmb:
@haampie, I believe this should address the concerns raised in https://github.com/ROCm/hipSOLVER/pull/252#issuecomment-2144871058.

library/src/cmake/suitesparse/FindCHOLMOD.cmake
```diff
@@ -0,0 +1,73 @@
...
+  if(NOT DEFINED CHOLMOD_LIBRARIES)
+    if(TARGET SuiteSparse::SuiteSparse_config)
+      set(CHOLMOD_LIBRARIES ${CHOLMOD_LIBRARY} SuiteSparse::SuiteSparse_config)
+    else()
+      set(CHOLMOD_LIBRARIES ${CHOLMOD_LIBRARY})
+    endif()
+  endif()
+
+  if(NOT TARGET SuiteSparse::CHOLMOD)
+    add_library(SuiteSparse::CHOLMOD INTERFACE IMPORTED)
```
jmachado-amd: I think there is a typo here, as [hipsolver consumes](https://github.com/ROCm/hipSOLVER/blob/13ebb6708a855238e344ca63670d3c6aad5af893/library/src/CMakeLists.txt#L146) the target [`CHOLMOD`](https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/26ababc7f3af725c5fb9168a1b94850eab74b666/CHOLMOD/CMakeLists.txt#L328) instead of the alias [`Suitesparse:CHOLMOD`](https://github.com/DrTimothyAldenDavis/SuiteSparse/blob/26ababc7f3af725c5fb9168a1b94850eab74b666/CMakeLists.txt#L360). Please, provide targets and aliases to match the ones originally defined by SuiteSparse.

Notwithstanding the previous syntactic issue, this code should work for many use cases. However, I disagree with the use of the `INTERFACE` property here. I understand that interface is meant to set requirements and define usage, and not to point to actual libraries on disk ([per the docs](https://cmake.org/cmake/help/latest/command/add_library.html#imported-libraries)). Thus, the canonical procedure would be to import those targets either as `STATIC` or `SHARED` libraries and at least set their `IMPORTED_LOCATION` (or `IMPORTED_LOCATION_<CONFIG>`) property (but you likely want to set both the `IMPORTED_LOCATION*` and `IMPORTED_IMPLIB*` properties for Windows users that might use this as well).

As is, cmake might not be able to properly handle exporting or installing hipsolver or clients that use those `Find*.cmake` files downstream. Moreover, Windows users will be able to compile hipsolver, but they might have issues running clients linking to hipsolver if the required SuiteSparse dlls are not in the path (as SuiteSparse's dlls won't be installed alongside hipsolver runtime).

library/src/cmake/suitesparse/FindSuiteSparse_config.cmake
```diff
@@ -0,0 +1,69 @@
...
+include(FindPackageHandleStandardArgs)
+find_package_handle_standard_args(SuiteSparse_config SUITESPARSE_CONFIG_INCLUDE_DIR SUITESPARSE_CONFIG_LIBRARY)
+
+if(SUITESPARSE_CONFIG_FOUND)
+  if(NOT DEFINED SUITESPARSE_CONFIG_LIBRARIES)
+    set(SUITESPARSE_CONFIG_LIBRARIES ${SUITESPARSE_CONFIG_LIBRARY})
+  endif()
+
+  if(NOT TARGET SuiteSparse::SuiteSparseConfig)
+    add_library(SuiteSparse::SuiteSparseConfig INTERFACE IMPORTED)
```
jmachado-amd: This creates the target `SuiteSparse::SuiteSparseConfig` which is an alias on the original SuiteSparse package. See analogous comment on file `FindCHOLMOD.cmake`.

jmachado-amd (changes requested):
The code reads very well and is well documented. Please, have a look at possible typos (i.e., defining targets that are originally aliases on SuiteSparse) and I will approve the PR right away.

That said, I am not completely familiar with the use case you have in your mind, so I also included comments on semantic issues related to imported and interface libraries, which may or may not be a problem here. Please, feel free to not implement those if they don't fit the intended use case. I am a bit wary about providing find modules to downstream users myself, so I tend to err on the "being way too careful" side on those things.

library/src/cmake/suitesparse/FindCHOLMOD.cmake
cgmb: > However, I disagree with the use of the INTERFACE property here. I understand that interface is meant to set requirements and define usage, and not to point to actual libraries on disk ([per the docs](https://cmake.org/cmake/help/latest/command/add_library.html#imported-libraries)). Thus, the canonical procedure would be to import those targets either as STATIC or SHARED libraries

I was following the example of FindLAPACK, but that one is not specifically a single library. It should probably be `UNKNOWN` rather than `INTERFACE`, [as is done for libraries like curl](https://github.com/Kitware/CMake/blob/v3.30.0/Modules/FindCURL.cmake#L213). I'll take a second look and update.

> As is, cmake might not be able to properly handle exporting or installing hipsolver or clients that use those Find*.cmake files downstream.

Good point.

jmachado-amd (approved):
Approving as I completely missed that you are already using the updated targets (`SuiteSparse::CHOLMOD`, etc.).

EdDAzevedo (approved):
Approve so that this is not a blocker.

PR #287, opened 2024-06-11
Added Get/SetDeterministicMode

tfalders:


library/src/amd_detail/hipsolver.cpp
```diff
@@ -607,6 +607,41 @@ catch(...)
...
 }
 
+hipsolverStatus_t hipsolverSetDeterministicMode(hipsolverHandle_t            handle,
+                                                hipsolverDeterministicMode_t mode)
+try
+{
+    if(!handle)
+        return HIPSOLVER_STATUS_NOT_INITIALIZED;
+
+    return hipsolver::rocblas2hip_status(rocblas_set_atomics_mode(
```
EdDAzevedo, line 617: Will this affect only rocblas or will it also need something like "rocsolver_set_atomics_mode()"?    Just checking.

EdDAzevedo (approved):

library/src/amd_detail/hipsolver.cpp
tfalders: To my knowledge, rocSOLVER doesn't use any non-deterministic programming patterns, so any non-determinism is either a bug or coming from rocBLAS. If we start using some non-deterministic patterns then we'll need to add something similar to rocSOLVER to be called by this method.

EdDAzevedo: It seems to me atomic operations are used in some source codes. 
Perhaps these are safe uses  on integer types  and so won't introduce non-deterministic behaviour?
Just checking.

src/lapack/roclapack_syevj_heevj.hpp:            atomicAdd(completed, 1);
src/lapack/roclapack_syevj_heevj.hpp:            atomicAdd(completed, 1);
src/refact/rocrefact_csrrf_splitlu.hpp:        atomicAdd(&(nzUarray[irow]), lnnzUrow);
src/refact/rocrefact_csrrf_splitlu.hpp:        atomicAdd(&(nzLarray[irow]), (lid == 0) ? 1 + lnnzLrow : lnnzLrow);
src/refact/rocrefact_csrrf_splitlu.hpp:        atomicAdd(&(nnzUrow[irow]), lnnzUrow);
src/include/lib_device_helpers.hpp:                _info_temp = atomicCAS(&_info, _info_temp, i + 1);
src/specialized/roclapack_getri_specialized_kernels.hpp:                _info_temp = atomicCAS(&_info, _info_temp, i + 1);

library/src/nvidia_detail/hipsolver_conversions.hpp
```diff
@@ -58,6 +58,12 @@ cusolverAlgMode_t hip2cuda_algmode(hipsolverAlgMode_t mode);
 
 hipsolverAlgMode_t cuda2hip_algmode(cusolverAlgMode_t mode);
 
+#if(!defined(CUDART_VERSION) || CUDART_VERSION < 12020)
```
cgmb: The `!defined(CUDART_VERSION) ||` is unnecessary. In preprocessor expressions, all identifiers that are not macros are treated as `0` ([GCC Manual §4.2.2](https://gcc.gnu.org/onlinedocs/gcc-3.0/cpp_4.html#SEC38)). If `CUDART_VERSION` is not defined, the expression `CUDART_VERSION < 12020` will be treated as `0 < 12020`.

library/include/internal/hipsolver-types.h
```diff
@@ -126,6 +125,12 @@ typedef enum
     HIPSOLVER_EIG_RANGE_I   = 223,
 } hipsolverEigRange_t;
 
+typedef enum
+{
+    HIPSOLVER_DETERMINISTIC_RESULTS           = 241,
+    HIPSOLVER_ALLOW_NON_DETERMINISTIC_RESULTS = 242,
+} hipsolverDeterministicMode_t;
```
cgmb, line 133: There have been requests to ensure compatibility between hipSOLVER and cuSOLVER all the way down to the values of enumerations. We should consider whether we want to adopt matching values immediately for new functionality, or keep following our existing pattern until there's a good time to change them all at once.

library/include/internal/hipsolver-types.h
```diff
@@ -14,8 +14,7 @@
 #endif
 
 #include <hip/hip_complex.h>
-#include <hip/hip_runtime_api.h>
-#include <hip/library_types.h>
+#include <hip/hip_runtime.h>
```
cgmb: Why include `hip/hip_runtime.h`? Isn't that primarily for HIP language code?

library/src/nvidia_detail/hipsolver_conversions.hpp
tfalders: I copied this macro from hipSPARSE, but if that part is unnecessary then I'll remove it.

library/include/internal/hipsolver-types.h
tfalders: Is that something we should wait for a major release to change, or can it be changed anytime?

library/include/internal/hipsolver-types.h
tfalders: I was attempting to fix the compile failure on cuda11. I figured CUDART_VERSION was not defined and `hip/hip_runtime.h` would include it, but this doesn't seem to have fixed the problem.

library/include/internal/hipsolver-types.h
cgmb: Changing it would be an ABI change (though not an API change). It would have to wait until at least ROCm 7.0.

library/src/nvidia_detail/hipsolver.cpp
```diff
@@ -89,6 +89,49 @@ catch(...)
...
+{
+#if(CUDART_VERSION >= 12020)
+    if(!handle)
+        return HIPSOLVER_STATUS_NOT_INITIALIZED;
+
+    return hipsolver::cuda2hip_status(cusolverSetDeterministicMode(
+        (cusolverDnHandle_t)handle, hipsolver::hip2cuda_deterministic(mode)));
+#else
+    return HIPSOLVER_STATUS_NOT_SUPPORTED;
+#endif
```
cgmb, line 104: We may want to ask around whether the preference is to return `HIPSOLVER_STATUS_NOT_SUPPORTED` or to drop the function from the API entirely. This is fine for now, but that's a discussion we should pencil in for later.

qjojo (approved):

library/include/internal/hipsolver-types.h
cgmb: If this is not needed, could we leave it as-is?

tfalders: I can try reverting it, sure. I'm unclear on which header defines CUDART_VERSION, so it's possible I might have to change it back to how it is now.

cgmb (approved):

PR #288, opened 2024-06-11
Reformat cmake files

cgmb:
There are a few manual touchups for indentation levels, but this is mostly just running:

    sed -Ei \
        -e 's/\( /\(/g' \
        -e 's/([^[:space:]]+) \)/\1\)/g' \
        -e 's/(if|elif|endif|set|list|file) \(/\1(/g' \
        -e 's/#(\w)/# \1/'

These transformations eliminate spacing after opening parentheses, before closing parentheses (if there was any text on the same line), between the keyword and the opening parentheses, and ensure there is a space after comment markers.

### DIFF
--- a/library/src/CMakeLists.txt
+++ b/library/src/CMakeLists.txt
@@ -22,7 +22,7 @@
 
 
 # ########################################################################
-# A helper function to prefix a source list of files with a common path into a new list(non-destructive)
+# A helper function to prefix a source list of files with a common path into a new list (non-destructive)
 # ########################################################################
 function(prepend_path prefix source_list_of_files return_list_of_files)
   foreach(file ${${source_list_of_files}})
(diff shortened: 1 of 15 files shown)


library/src/CMakeLists.txt
```diff
@@ -22,26 +22,26 @@
 
 
 # ########################################################################
-# A helper function to prefix a source list of files with a common path into a new list (non-destructive)
+# A helper function to prefix a source list of files with a common path into a new list(non-destructive)
```
tfalders: Looks like this comment got caught in the automatic replacement

tfalders (approved):
Awesome! Thanks a bunch for this

library/src/CMakeLists.txt
cgmb: Good catch.

jmachado-amd (approved):
That is quite an aesthetic improvement, thanks for doing it Cory!

PR #289, opened 2024-06-12
Drop debug postfix

cgmb:
Users that wish to continue building with the debug postfix enabled can pass `-DCMAKE_DEBUG_POSTFIX="-d"` to CMake to restore the postfix for the library. I have not tested, but I believe a postfix for executables may be added using `-DCMAKE_EXECUTABLE_SUFFIX="-d"` on Linux or `-DCMAKE_EXECUTABLE_SUFFIX="-d.exe"` on Windows.

### DIFF
--- a/library/src/CMakeLists.txt
+++ b/library/src/CMakeLists.txt
@@ -24,24 +24,24 @@
 # ########################################################################
 # A helper function to prefix a source list of files with a common path into a new list (non-destructive)
 # ########################################################################
-function( prepend_path prefix source_list_of_files return_list_of_files )
-  foreach( file ${${source_list_of_files}} )
-    if(IS_ABSOLUTE ${file} )
-      list( APPEND new_list ${file} )
-    else( )
-      list( APPEND new_list ${prefix}/${file} )
-    endif( )
-  endforeach( )
-  set( ${return_list_of_files} ${new_list} PARENT_SCOPE )
-endfunction( )
+function(prepend_path prefix source_list_of_files return_list_of_files)
+  foreach(file ${${source_list_of_files}})
+    if(IS_ABSOLUTE ${file})
+      list(APPEND new_list ${file})
+    else()
+      list(APPEND new_list ${prefix}/${file})
+    endif()
+  endforeach()
+  set(${return_list_of_files} ${new_list} PARENT_SCOPE)
+endfunction()
 
 # ########################################################################
 # Main
 # ########################################################################
-prepend_path( ".." hipsolver_headers_public relative_hipsolver_headers_public )
+prepend_path(".." hipsolver_headers_public relative_hipsolver_headers_public)
 
-if( NOT USE_CUDA )
-  set( hipsolver_source
+if(NOT USE_CUDA)
+  set(hipsolver_source
     "${CMAKE_CURRENT_SOURCE_DIR}/amd_detail/hipsolver_conversions.cpp"
     "${CMAKE_CURRENT_SOURCE_DIR}/amd_detail/hipsolver.cpp"
     "${CMAKE_CURRENT_SOURCE_DIR}/amd_detail/hipsolver_dense.cpp"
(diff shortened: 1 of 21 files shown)


tfalders (approved):

jmachado-amd (changes requested):
Although I have no objections with removing the postfix "-d" from debug executables, the PR description caught my attention:

> Users that wish to continue building with the debug postfix enabled can pass -DCMAKE_DEBUG_POSTFIX="-d" when building hipSOLVER.

I understand that `DCMAKE_DEBUG_POSTFIX` only changes the name of non-executable targets, while all that I see is the PR touching code that set the name of executables.

https://cmake.org/cmake/help/latest/variable/CMAKE_CONFIG_POSTFIX.html

Other than this apparent contradiction, everything looks good. Let me know if the intent of the PR is really to deprecate the postfix "-d" on executables, or if I got anything wrong, and I will approve it right away.

cgmb:
> I understand that `DCMAKE_DEBUG_POSTFIX` only changes the name of non-executable targets, while all that I see is the PR touching code that set the name of executables.

I didn't notice that. Thanks for pointing it out! I suspect that the CMake reasoning is that setting the debug postfix on executable targets is weird and unusual, which is why this mechanism only sets it on library targets. Although, I think you could use `CMAKE_EXECUTABLE_SUFFIX` for that if you really wanted to.

For what it's worth, [hipFFT removed it ten months ago](https://github.com/ROCm/hipFFT/commit/fcff4338f3be7bb776ef98a4bc29f2a9a5813068).

cgmb:
I'll update the description and add a changelog entry.

library/src/CMakeLists.txt
```diff
@@ -92,7 +92,7 @@ if( WIN32 )
     if( ${CMAKE_BUILD_TYPE} MATCHES "Debug" )
       add_custom_command( TARGET hipsolver
         POST_BUILD
-        COMMAND ${CMAKE_COMMAND} -E copy "${PROJECT_BINARY_DIR}/staging/hipsolver-d.pdb" "${PROJECT_BINARY_DIR}/clients/staging/hipsolver-d.pdb"
+        COMMAND ${CMAKE_COMMAND} -E copy "${PROJECT_BINARY_DIR}/staging/hipsolver.pdb" "${PROJECT_BINARY_DIR}/clients/staging/hipsolver.pdb"
```
jmachado-amd: I am wondering whether this copy can fail if someone sets `CMAKE_DEBUG_POSTFIX` or `CMAKE_EXECUTABLE_SUFFIX` for a build.

Mor precisely, I don't know whether setting `CMAKE_DEBUG_POSTFIX` or `CMAKE_EXECUTABLE_SUFFIX` would change the names of `pdb`s (although I would assume that those should fallow the name assigned to the .libs), and if they do, whether it is in a self-consistent manner. Given that `CMAKE_EXECUTABLE_SUFFIX` also includes the `.exe` in Windows, I would suspect that non-obvious interactions may happen when those two are mixed.

jmachado-amd (approved):
Everything looks good, and as the intent is clear then it is all fine with me.

I personally don't like the "-d" postfix, but since it is there, I just wonder whether anyone might be depending on those names. Anyway, I agree that users can revert to the old behavior by setting `CMAKE_DEBUG_SUFFIX` and `CMAKE_EXECUTABLE_SUFFIX`, but I've just added a comment regarding the only point I am not so sure about.

library/src/CMakeLists.txt
cgmb: It probably will fail. The code that copies files around on Windows is ugly and brittle in many ways.

Just for a bit of background, this operation is performed in the build code because our Windows CI only copies the files in `${PROJECT_BINARY_DIR}/clients/staging/` from the build machine to the test machine. IMO, we should find a better way to handle that. For example, running the tests on the library install tree rather than on the build tree.

Maybe we could add an option so that these copies can be disabled, and ease the transition to an eventual future where they do not exist.

cgmb: I've updated the code to use the `TARGET_PDB_FILE_NAME` generator expression, which should contain the correct name in either case.

PR #302, opened 2024-07-17
Manage sparse dependencies using dlopen

tfalders:
Companion to https://github.com/ROCm/rocSOLVER/pull/634

Remaining work:
1. Ensure .so and .dll filenames are correct and the library works properly on other OSes.
2. ~~Add rocSPARSE and SuiteSparse as recommended dependencies.~~
3. ~~Have rocSPARSE and SuiteSparse be build-time dependencies for the static lib.~~


clients/gtest/csrlsvchol_gtest.cpp
```diff
@@ -66,9 +66,15 @@ template <bool HOST>
 class CSRLSVCHOL_BASE : public ::TestWithParam<csrlsvchol_tuple>
 {
 protected:
-    CSRLSVCHOL_BASE() {}
-    virtual void SetUp() {}
-    virtual void TearDown() {}
+    void SetUp() override
+    {
+        if(hipsolverSpCreate(nullptr) == HIPSOLVER_STATUS_NOT_SUPPORTED)
+            GTEST_SKIP() << "Sparse functionality is not enabled";
```
qjojo: I think this error message should explicitly say that hipSOLVER hasn't been compiled with sparse support, just saying "enabled" might make users think there's a switch or environment variable they need to set to enable it.

qjojo: Reading further on it looks like this is as specific as we can be right here, but I still think more specific error messages would be a good idea since there are two distinct failure states for sparse not compiled in vs sparse couldn't load.

tfalders: I can change it to "Sparse dependencies could not be loaded", or something similar

CMakeLists.txt
```diff
@@ -225,20 +232,24 @@ if(NOT USE_CUDA)
...
     rocm_package_add_rpm_dependencies(STATIC_DEPENDS "rocsparse-static-devel >= ${rocsparse_minimum}")
     rocm_package_add_deb_dependencies(STATIC_DEPENDS "rocsparse-static-dev >= ${rocsparse_minimum}")
 
-    if(SYSTEM_OS STREQUAL "centos" OR SYSTEM_OS STREQUAL "rhel" OR SYSTEM_OS STREQUAL "mariner")
-      list(APPEND hipsolver_pkgdeps "suitesparse")
-    elseif(SYSTEM_OS STREQUAL "ubuntu" AND SYSTEM_OS_VERSION VERSION_GREATER_EQUAL "24.04")
-      list(APPEND hipsolver_pkgdeps "libcholmod5" "libsuitesparseconfig7")
-    else()
-      list(APPEND hipsolver_pkgdeps "libcholmod3" "libsuitesparseconfig5")
-    endif()
```
cgmb, line 243: This part is still relevant.

cgmb: Oh, wait, it just moved. My mistake.

clients/gtest/csrlsvchol_gtest.cpp
```diff
@@ -66,6 +66,11 @@ template <bool HOST>
 class CSRLSVCHOL_BASE : public ::TestWithParam<csrlsvchol_tuple>
 {
 protected:
+    void SetUp() override
+    {
+        if(hipsolverSpCreate(nullptr) == HIPSOLVER_STATUS_NOT_SUPPORTED)
+            GTEST_SKIP() << "Sparse dependencies could not be loaded";
```
cgmb, line 72: This is fine for now. We may at some point want to add a method to require the sparse tests, to ensure that we don't introduce a bug that breaks loading. Maybe we can check if `HIPSOLVER_SPARSE_TESTS_REQUIRED=1` is set in the environment.

library/src/amd_detail/dlopen/cholmod.cpp
```diff
@@ -0,0 +1,99 @@
...
+static bool load_cholmod()
+{
+#ifdef _WIN32
+    // Library users will need to call SetErrorMode(SEM_FAILCRITICALERRORS) if
+    // they wish to avoid an error message box when this library is not found.
+    // The call is not done by hipSOLVER directly, as it is not thread-safe and
+    // will affect the global state of the program.
+    void* handle = LoadLibraryW(L"cholmod.dll");
+#else
+    void* handle = dlopen("libcholmod.so", RTLD_NOW | RTLD_LOCAL);
```
cgmb: One potential improvement to consider. On Debian and Ubuntu, the `libcholmod.so` symlink is provided by the `-dev` package so that multiple versions of the library with different major versions can be installed at the same time. I would suggest that we try loading `libcholmod.so.5`, `libcholmod.so.4` or `libcholmod.so.3` and _maybe_ try loading `libcholmod.so` if those all fail. I should probably do the same thing to support both `librocsparse.so.0` and `librocsparse.so.1` in rocsolver.

library/src/amd_detail/dlopen/rocsparse.cpp
```diff
@@ -0,0 +1,73 @@
...
+fp_rocsparse_get_mat_index_base g_rocsparse_get_mat_index_base;
+
+static bool load_rocsparse()
+{
+#ifdef _WIN32
+    // Library users will need to call SetErrorMode(SEM_FAILCRITICALERRORS) if
+    // they wish to avoid an error message box when this library is not found.
+    // The call is not done by hipSOLVER directly, as it is not thread-safe and
+    // will affect the global state of the program.
+    void* handle = LoadLibraryW(L"librocsparse.dll");
```
cgmb: It should be `rocsparse.dll`. Sorry, you copied that mistake from me.

library/src/amd_detail/dlopen/load_function.hpp
```diff
@@ -0,0 +1,53 @@
...
+ *
+ *
+ * ************************************************************************ */
+
+#include "lib_macros.hpp"
+
+#include <iostream>
+
+#ifdef _WIN32
+#include <libloaderapi.h>
```
cgmb: It seems this should be `#include <windows.h>`.

library/src/amd_detail/dlopen/load_function.hpp
```diff
@@ -0,0 +1,53 @@
...
+#else
+#include <dlfcn.h>
+#endif /* _WIN32 */
+
+HIPSOLVER_BEGIN_NAMESPACE
+
+template <typename Fn>
+static bool load_function(void* handle, const char* symbol, Fn& fn)
```
cgmb: I would avoid `static` for templates in headers, as that will prevent duplicate instantiations from being combined during linking.

cgmb (approved):
LGTM. Please ensure that the tests all pass CI (with no skips) before merging.

cgmb:
> LGTM. Please ensure that the tests all pass CI (with no skips) before merging.

I forgot that we didn't have sparse enabled on Windows before. The criteria is just that there should be no skipping on Linux on the CI.

qjojo (approved):

tfalders:
Other than on Windows, I see no skips and the tests appear to be passing. I'll perform the merge.

PR #318, opened 2024-08-13
Add api for DnXpotrf and DnXpotrs

qjojo:


jzuniga-amd (approved):
LGTM

library/src/nvidia_detail/hipsolver_dense64.cpp
```diff
@@ -192,4 +192,69 @@ catch(...)
...
+                                               size_t*             lworkOnHost)
+try
+{
+    if(!handle)
+        return HIPSOLVER_STATUS_NOT_INITIALIZED;
+    if(!params)
+        return HIPSOLVER_STATUS_INVALID_VALUE;
+
+    return hipsolver::cuda2hip_status(cusolverDnXpotrf_bufferSize(
+        handle, params, uplo, n, dataTypeA, A, lda, computeType, lworkOnDevice, lworkOnHost));
```
jmachado-amd: I am still not completely familiar with hipSOLVER but I thought that `handle` and `params` were meant to be cast to types `cusolverDnHandle_t` and `cusolverDnParams_t`. Also, `cusolverDnXpotrf_bufferSize` should probably expect the result of `hipsolver::hip2cuda_operation(uplo)` (or something similar) instead of a `uplo` (which has type `hipsolverFillMode_t`).

jmachado-amd (commented):
It looks like build is not passing in CI (cuda), please have a look at the types of the parameters of methods `hipsolverDnXpotrf_bufferSize` and `hipsolverDnXpotrf` (in file `nvidia_detail/hipsolver_dense.hpp`) as some probably need to be cast to proper cusolver types.

jmachado-amd (approved):
Looks good!

jzuniga-amd:
CI cuda is passing now. Merging this to complete 6.3 dev cycle process.

PR #334, opened 2024-10-24
Disable sparse functions by default when building a static library

qjojo:
Disable sparse functions by default when building a static library.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -109,7 +109,7 @@ option(BUILD_CODE_COVERAGE "Build hipSOLVER with code coverage enabled" OFF)
 option(BUILD_HIPBLAS_TESTS "Build additional tests to ensure hipBLAS and hipSOLVER are compatible (requires installed hipBLAS)" OFF)
 option(BUILD_HIPSPARSE_TESTS "Build additional tests to cover sparse functionality (requires installed hipSPARSE)" ON)
 # BUILD_SHARED_LIBS is a cmake built-in; we make it an explicit option such that it shows in cmake-gui
-option(BUILD_WITH_SPARSE "Build hipSOLVER with sparse functionality available at build time (requires installed dependencies)" ON)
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -104,11 +104,7 @@ list(APPEND CMAKE_MODULE_PATH ${CMAKE_CURRENT_SOURCE_DIR}/cmake ${ROCM_PATH}/lib
 list(APPEND CMAKE_PREFIX_PATH ${ROCM_PATH}/lib/cmake/hip /opt/rocm /opt/rocm/llvm /opt/rocm/hip)
 
 option(BUILD_SHARED_LIBS "Build hipSOLVER as a shared library" ON)
-if(BUILD_SHARED_LIBS)
-  set(BUILD_WITH_SPARSE_DEFAULT OFF)
-else()
-  set(BUILD_WITH_SPARSE_DEFAULT ON)
-endif()
+set(BUILD_WITH_SPARSE_DEFAULT OFF)
```
tfalders: You can probably remove BUILD_WITH_SPARSE_DEFAULT entirely. I think we only added it to manage the different defaults between the shared and static lib.

Also, do you not need to modify load_function, load_rocsparse, and load_cholmod the same way as in rocSOLVER?

jzuniga-amd (approved):

qjojo:
Added the include guards from rocSOLVER and the HIPSOLVER_STATIC_LIB define

CMakeLists.txt
tfalders: We ought to remove it from line 112 as well. I'll approve once that's done.

qjojo:
The test failure is a segfault on sygvdx running on CUDA, I don't think that has anything to do with these changes.

cgmb (approved):

tfalders (approved):

jmachado-amd:
> The test failure is a segfault on sygvdx running on CUDA, I don't think that has anything to do with these changes.

Yes, the segfaults on CUDA's sygvdx are happening for some time now.

PR #335, opened 2024-10-28
merge release-staging/rocm-rel-6.3 back into develop

jzuniga-amd:
we need to have all commits/changes in the release branch back into develop/staging for CI after branching.

(note: I also updated the unreleased part of the CHANGELOG to reflect the new format)


tfalders (approved):
Looks good. Capitalization on "Issues" and "Changes" looks different from the new format. They should probably be uncapitalized.

PR #338, opened 2024-11-13
Add sparse QR solver functions (csrlsvqr)

qjojo:
This adds hipsolverSpXcsrlsvqr for all scalar types. To accomplish this it also:

- Loads additional rocsparse functions (rocsparse_*csr2dense, rocsparse_create_handle, rocsparse_destroy_handle)
- Adds a field containing a rocsparse handle to hipsolverSpHandle_t
- Adds a shim to create complex test cases from real ones.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -97,9 +97,9 @@ Full documentation for hipSOLVER is available at the [hipSOLVER Documentation](h
 
 * Added compatibility API with hipsolverSp prefix
 * Added compatibility-only functions
-  * csrlsvchol
-    * hipsolverSpScsrlsvcholHost, hipsolverSpDcsrlsvcholHost
-    * hipsolverSpScsrlsvchol, hipsolverSpDcsrlsvchol
+  * `csrlsvchol`
+    * `hipsolverSpScsrlsvcholHost`, `hipsolverSpDcsrlsvcholHost`
+    * `hipsolverSpScsrlsvchol`, `hipsolverSpDcsrlsvchol`
 * Added rocSPARSE and SuiteSparse as optional dependencies to hipSOLVER (rocSOLVER backend only). Use the `BUILD_WITH_SPARSE` CMake option to enable
   functionality for the hipsolverSp API (on by default).
 * Added hipSPARSE as an optional dependency to hipsolver-test. Use the `BUILD_WITH_SPARSE` CMake option to enable tests of the hipsolverSp API (on by default).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -97,9 +97,9 @@ Full documentation for hipSOLVER is available at the [hipSOLVER Documentation](h
 
 * Added compatibility API with hipsolverSp prefix
 * Added compatibility-only functions
-  * `csrlsvchol`
-    * `hipsolverSpScsrlsvcholHost`, `hipsolverSpDcsrlsvcholHost`
-    * `hipsolverSpScsrlsvchol`, `hipsolverSpDcsrlsvchol`
+  * csrlsvchol
+    * hipsolverSpScsrlsvcholHost, hipsolverSpDcsrlsvcholHost
+    * hipsolverSpScsrlsvchol, hipsolverSpDcsrlsvchol
 * Added rocSPARSE and SuiteSparse as optional dependencies to hipSOLVER (rocSOLVER backend only). Use the `BUILD_WITH_SPARSE` CMake option to enable
   functionality for the hipsolverSp API (on by default).
 * Added hipSPARSE as an optional dependency to hipsolver-test. Use the `BUILD_WITH_SPARSE` CMake option to enable tests of the hipsolverSp API (on by default).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ Full documentation for hipSOLVER is available at the [hipSOLVER Documentation](h
 ## (Unreleased) hipSOLVER
 
 ### Added
+* Added compatibility-only functions
+  * csrlsvqr
+    * hipsolverSpScsrlsvqr, hipsolverSpDcsrlsvqr, hipsolverSpCcsrlsvqr, hipsolverSpZcsrlsvqr
 ### Changed
 ### Removed
 ### Optimized
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Full documentation for hipSOLVER is available at the [hipSOLVER Documentation](h
 * Added compatibility-only functions
   * csrlsvqr
     * hipsolverSpScsrlsvqr, hipsolverSpDcsrlsvqr, hipsolverSpCcsrlsvqr, hipsolverSpZcsrlsvqr
+
 ### Changed
 ### Removed
 ### Optimized
--- a/clients/include/testing_csrlsvqr.hpp
+++ b/clients/include/testing_csrlsvqr.hpp
@@ -233,8 +233,7 @@ void csrlsvqr_getError(hipsolverSpHandle_t       handle,
     err      = norm_error('I', n, 1, n, hX[0], hXRes[0]);
     *max_err = err > *max_err ? err : *max_err;
 
-    // TODO: Add non-positive definite test matrices
-    // also check info for singularities
+    // TODO: Add singular matrices and check info
     err = 0;
     EXPECT_EQ(hSingularity[0][0], -1);
     if(hSingularity[0][0] != -1)
--- a/clients/include/testing_csrlsvqr.hpp
+++ b/clients/include/testing_csrlsvqr.hpp
@@ -1,0 +1,639 @@
+/* ************************************************************************
+ * Copyright (C) 2024 Advanced Micro Devices, Inc.
+ * ************************************************************************ */
+
+#pragma once
+
+#include "clientcommon.hpp"
+#include "hipsolverSp.hpp"
+
+template <bool HOST, typename T>
+void csrlsvqr_checkBadArgs(hipsolverSpHandle_t       handle,
+                           const int                 n,
+                           const int                 nnzA,
+                           const hipsparseMatDescr_t descrA,
+                           int*                      ptrA,
+                           int*                      indA,
+                           T                         valA,
+                           T                         B,
+                           T                         X,
+                           int*                      singularity)
+{
+    // handle
+    EXPECT_ROCBLAS_STATUS(
+        hipsolver_csrlsvqr(
+            HOST, nullptr, n, nnzA, descrA, valA, ptrA, indA, B, 0, 0, X, singularity),
+        HIPSOLVER_STATUS_NOT_INITIALIZED);
+
+    // values
+    // N/A
+
+#if defined(__HIP_PLATFORM_HCC__) || defined(__HIP_PLATFORM_AMD__)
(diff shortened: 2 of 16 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,9 @@ Full documentation for hipSOLVER is available at the [hipSOLVER Documentation](h
 ## (Unreleased) hipSOLVER
 
 ### Added
+* Added compatibility-only functions
+  * csrlsvqr
+    * hipsolverSpScsrlsvqr, hipsolverSpDcsrlsvqr, hipsolverSpCcsrlsvqr, hipsolverSpZcsrlsvqr
```
amd-jnovotny, line 10:
```suggestion
  * `csrlsvqr`
    * `hipsolverSpScsrlsvqr`, `hipsolverSpDcsrlsvqr`, `hipsolverSpCcsrlsvqr`, `hipsolverSpZcsrlsvqr`
```

library/src/amd_detail/hipsolver_sparse.cpp
```diff
@@ -340,23 +341,33 @@ try
...
         rocblas_destroy_handle(sp->handle);
+        rocsparse_destroy_handle(sp->sphandle);
         delete sp;
         return hipsolver::rocblas2hip_status(status);
     }
 
     if(cholmod_start(&sp->c_handle) != TRUE)
     {
         rocblas_destroy_handle(sp->handle);
+        rocsparse_destroy_handle(sp->sphandle);
```
tfalders, line 370: I believe you also need to destroy the rocSPARSE handle in hipsolverSpDestroy.

library/src/amd_detail/hipsolver_sparse.cpp
```diff
@@ -1074,4 +1085,275 @@ catch(...)
...
+        return HIPSOLVER_STATUS_NOT_INITIALIZED;
+    if(n < 0 || nnz < 0)
+        return HIPSOLVER_STATUS_INVALID_VALUE;
+    if(!descrA)
+        return HIPSOLVER_STATUS_INVALID_VALUE;
+    if(!csrRowPtr || !csrColInd || !csrVal || !descrA)
+        return HIPSOLVER_STATUS_INVALID_VALUE;
+    if(!b || !x || !singularity)
+        return HIPSOLVER_STATUS_INVALID_VALUE;
+
```
tfalders, line 1116: I would suggest we keep the argument check on reorder, even though we're not using it at the moment. (Also applies to the other three functions)

library/src/amd_detail/hipsolver_sparse.cpp
```diff
@@ -1074,4 +1085,275 @@ catch(...)
...
+        sp->sphandle, n, n, (rocsparse_mat_descr)descrA, csrVal, csrRowPtr, csrColInd, denseA, n);
+
+    int* info;
+    CHECK_HIP_ERROR(hipMalloc(&info, sizeof(int)));
+
+    rocblas_status st
+        = rocsolver_sgels(sp->handle, rocblas_operation_none, n, n, 1, denseA, n, x, n, info);
+
+    CHECK_HIP_ERROR(hipFree(denseA));
+    CHECK_HIP_ERROR(hipFree(info));
```
tfalders, line 1147: Before deleting info, we can hipMemcpy it to singularity and then decrement singularity by 1. This should set singularity corrently. (Also applies to the other functions)

library/src/nvidia_detail/hipsolver_sparse.cpp
```diff
@@ -373,4 +373,153 @@ catch(...)
...
+    if(!handle)
+        return HIPSOLVER_STATUS_NOT_INITIALIZED;
+    if(!descrA)
+        return HIPSOLVER_STATUS_INVALID_VALUE;
+
+    return hipsolver::cuda2hip_status(cusolverSpCcsrlsvqr((cusolverSpHandle_t)handle,
+                                                          n,
+                                                          nnzA,
+                                                          (cusparseMatDescr_t)descrA,
+                                                          csrVal,
```
tfalders: If I remember correctly, hipFloatComplex* and hipDoubleComplex* need to be cast to cuComplex* and cuDoubleComplex* when passing to cuSOLVER.

library/src/amd_detail/hipsolver_sparse.cpp
```diff
@@ -1074,4 +1085,275 @@ catch(...)
...
+
+    hipsolverSpHandle* sp = (hipsolverSpHandle*)handle;
+    *singularity          = -1;
+
+    // set up B
+    CHECK_HIP_ERROR(hipMemcpy((void*)x, b, sizeof(float) * n, hipMemcpyDeviceToDevice));
+
+    // convert A to dense matrix
+    float* denseA;
+    CHECK_HIP_ERROR(hipMalloc(&denseA, sizeof(float) * n * n));
```
tfalders, line 1132: Ideally, this would be allocated in the hipsolverSp handle, but converting to dense is a temporary workaround so I'm perfectly fine with keeping it here.

tfalders (changes requested):
Looks like you didn't even need my help to transfer the implementation over. Nice job!

I just have a few comments to be taken care of. We should also update docs/howto/usage.rst but I can handle that for you.

tfalders:
SYGVDX is segfaulting on the cuda side, which will prevent us from seeing if the sparse QR tests pass. We might want to add these tests to known_bugs or disable them so that we can proceed.

tfalders:
We might also want to disable support for the complex precisions, at least until we know if the final implementation will support them.

tfalders:
@qjojo I've opened a PR on your local fork with my suggested changes

docs/howto/usage.rst
```diff
@@ -115,16 +115,20 @@ Unsupported methods
   * :ref:`hipsolverSpXcsrlsvcholHost <sparse_csrlsvcholHost>` with `reorder = 1`
   * :ref:`hipsolverSpXcsrlsvchol <sparse_csrlsvchol>` with `reorder = 1`
 
+- The functions :ref:`hipsolverSpScsrlsvqr <sparse_csrlsvqr>` are currently implemented by converting the sparse input matrix to a dense
```
amd-jnovotny:
```suggestion
- The function :ref:`hipsolverSpScsrlsvqr <sparse_csrlsvqr>` is currently implemented by converting the sparse input matrix to a dense
```

docs/howto/usage.rst
```diff
@@ -134,6 +138,13 @@ Performance implications of the hipsolverSp API
   (:ref:`hipsolverSpXcsrlsvchol <sparse_csrlsvchol>` may perform slower and will require more memory usage than
   :ref:`hipsolverSpXcsrlsvcholHost <sparse_csrlsvcholHost>`.)
 
+- The functions :ref:`hipsolverSpScsrlsvqr <sparse_csrlsvqr>` are currently implemented by converting the sparse input matrix to a dense
+  matrix, and then running the dense factorization and linear solver on the result. This may result in slower-than-expected performance and
+  significant memory usage for large matrices.
```
amd-jnovotny:
```suggestion
- The function :ref:`hipsolverSpScsrlsvqr <sparse_csrlsvqr>` is currently implemented by converting the sparse input matrix to a dense
  matrix, and then running the dense factorization and linear solver on the result. This might result in slower-than-expected performance and
  significant memory usage for large matrices.
```

amd-jnovotny (commented):
Just a couple of minor editorial suggestions for the docs changes.

tfalders:
@qjojo I've opened a second PR on your fork because the first one didn't fix the segfault with cuda testing

tfalders (approved):
LGTM! (Assuming the cuda test passes)

CHANGELOG.md
```diff
@@ -93,9 +97,9 @@ Full documentation for hipSOLVER is available at the [hipSOLVER Documentation](h
 
 * Added compatibility API with hipsolverSp prefix
 * Added compatibility-only functions
-  * csrlsvchol
-    * hipsolverSpScsrlsvcholHost, hipsolverSpDcsrlsvcholHost
-    * hipsolverSpScsrlsvchol, hipsolverSpDcsrlsvchol
+  * `csrlsvchol`
+    * `hipsolverSpScsrlsvcholHost`, `hipsolverSpDcsrlsvcholHost`
+    * `hipsolverSpScsrlsvchol`, `hipsolverSpDcsrlsvchol`
```
jzuniga-amd: I wonder why these functions are highlighted with `` 
We are not doing that with any other function

clients/include/testing_csrlsvqr.hpp
```diff
@@ -0,0 +1,639 @@
...
+
+    // compare computed results with original result
+    double err;
+    *max_err = 0;
+
+    err      = norm_error('I', n, 1, n, hX[0], hXRes[0]);
+    *max_err = err > *max_err ? err : *max_err;
+
+    // TODO: Add non-positive definite test matrices
+    // also check info for singularities
```
jzuniga-amd: Here what we need to add in the future is singular matrices.
The TODO comment needs to be updated.

docs/howto/usage.rst
```diff
@@ -115,25 +115,36 @@ Unsupported methods
   * :ref:`hipsolverSpXcsrlsvcholHost <sparse_csrlsvcholHost>` with `reorder = 1`
   * :ref:`hipsolverSpXcsrlsvchol <sparse_csrlsvchol>` with `reorder = 1`
 
+- The function :ref:`hipsolverSpScsrlsvqr <sparse_csrlsvqr>` is currently implemented by converting the sparse input matrix to a dense
+  matrix, and therefore do not support any reordering method.
```
jzuniga-amd: we should also mention that the HOST path is not supported at the moment

library/src/amd_detail/hipsolver_sparse.cpp
```diff
@@ -1074,4 +1086,299 @@ catch(...)
...
+hipsolverStatus_t hipsolverSpDcsrlsvqr(hipsolverSpHandle_t       handle,
+                                       int                       n,
+                                       int                       nnz,
+                                       const hipsparseMatDescr_t descrA,
+                                       const double*             csrVal,
+                                       const int*                csrRowPtr,
+                                       const int*                csrColInd,
+                                       const double*             b,
+                                       double                    tolerance,
+                                       int                       reorder,
```
jzuniga-amd, line 1165: `tolerance` and `reorder` are ignored because we solve with a dense matrix (with gels). 
The exact form of the QR factors doesn't matter when the output is only the solution vector X, however, a non-zero tolerance could give different singularity results for rocm and cuda for the same input matrix. 

This should be at least mentioned in the documentation (Maybe in the "some considerations when using SP API" section)

We could also have an internal `rocsolver_gels` for use by hipsolver that checks singularity within a given tolerance, but it is maybe too much if this is not going to be the final csrlsvqr implementation

jzuniga-amd (commented):
Looks good. I think we only need a couple more comments in the documentation. See below:

library/src/amd_detail/hipsolver_sparse.cpp
tfalders: On second thought, I've decided to comment out the check on reorder. That way, we can say reorder is not referenced by the rocSOLVER backend, and having the comment there will remind us to reenable it once we using it.

jzuniga-amd (approved):
LGTM

PR #360, opened 2025-02-06
Fix static build

jmachado-amd:
Hard-code `-I${HIP_INCLUDE_DIRS}` as a compile option of target `example-c-basic` to fix the static build.

### DIFF
--- a/clients/samples/CMakeLists.txt
+++ b/clients/samples/CMakeLists.txt
@@ -1,5 +1,5 @@
 # ########################################################################
-# Copyright (C) 2016-2024 Advanced Micro Devices, Inc. All rights reserved.
+# Copyright (C) 2016-2025 Advanced Micro Devices, Inc. All rights reserved.
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


clients/samples/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ add_executable(example-cpp-basic example_basic.cpp)
 # We test for C99 compatibility in the example-c.c test
 set_source_files_properties(example_basic.c PROPERTIES LANGUAGE CXX)
 set_source_files_properties(example_basic.c PROPERTIES COMPILE_FLAGS "-xc -std=c99")
+target_compile_options(example-c-basic PRIVATE -I${HIP_INCLUDE_DIRS})
```
EdDAzevedo, line 29: I wonder perhaps this option -I${HIP_INCLUDE_DIRS} is safe and harmless  for other examples as well.  Just a thought.

EdDAzevedo (approved):

cgmb (changes requested):
-1. We are going to need to fix the actual root cause because a bug in `hip::host` is certain to affect other users besides hipsolver. And if we fix the root cause, there will be no need for this workaround.

jmachado-amd:
> -1. We are going to need to fix the actual root cause because a bug in hip::host is certain to affect other users besides hipsolver. And if we fix the root cause, there will be no need for this workaround.

@cgmb, I am afraid that this is not a bug, as its stems from a series of design choices made in the hip runtime, i.e., the issue at hand is a result of their intended behaviour.

I agree that this is surely a major pain to all libraries, and that it should have been communicated to everyone; but the one thing that we can and have to fix is to make sure that we get the memo the next time something similar happens.

clients/samples/CMakeLists.txt
jmachado-amd: The thing is that, unfortunately, this change should not have been necessary. The include directory pointed by `${HIP_INCLUDE_DIRS}` is being directly, and properly, set on target `example-c-basic` by the `target_include_directories` in line `36`; as well as being also indirectly set by the nearby `target_link_libraries` in some code paths. 

The problem lies in the fact that current static libraries builds are purposefully ignoring the hip runtime include dirs set in the proper way.

ammallya:

PR #361, opened 2025-02-10
Use amdclang++ instead of hipcc in static builds

jmachado-amd:
This updates the `install.sh` script to use `amdclang++` instead of `hipcc` for static builds, as the latter is deprecated.

(Refactoring the `install.sh` script is not the objective of this PR.)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -53,7 +53,7 @@ function display_help()
   echo "    [--hipsparse-path] Set specific path to custom built hipsparse"
   echo "    [--static] Create static library instead of shared library"
   echo "    [--codecoverage] Build with code coverage profiling enabled, excluding release mode."
-  echo "    [--address-sanitizer] Build with address sanitizer enabled. Uses hipcc to compile"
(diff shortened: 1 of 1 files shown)


cgmb (commented):
There's another use of `hipcc` in install.sh. This is a good change, but could you do a quick grep and ensure that we change all references?

install.sh
```diff
@@ -609,8 +609,8 @@ fi
       exit 1
     fi
     cmake_common_options+=("-DBUILD_SHARED_LIBS=OFF")
-    compiler="${rocm_path}/bin/hipcc" #force hipcc for static libs, g++ doesn't work
-    printf "Forcing compiler to hipcc for static library.\n"
+    compiler="${rocm_path}/bin/amdclang++" #force hipcc for static libs, g++ doesn't work
```
tfalders: Minor comment: The comment still mentions hipcc

tfalders (approved):

qjojo (approved):

install.sh
jmachado-amd: Yes, I missed this and that `hipcc` was still also used for asan builds.

PR #364, opened 2025-02-12
Use proper include path for cholmod.h.

stellaraccident:
It appears that the `SuiteSparse::CHOLMOD` CMake library advertises its include directory as `$PREFIX/include/suitesparse`, making the proper way to include it `#include <cholmod.h>`. When installed in the system `/usr` prefix, the `suitesparse/cholmod.h` path works by accident. When using a custom built suitesparse at a custom location, this is not so.

I believe that this change is the correct way to do it for all configurations, based on how the SuiteSparse library is set up.

### DIFF
--- a/library/src/amd_detail/dlopen/cholmod.hpp
+++ b/library/src/amd_detail/dlopen/cholmod.hpp
@@ -24,7 +24,7 @@
 #include "lib_macros.hpp"
 
 #ifdef HAVE_ROCSPARSE
-#include <suitesparse/cholmod.h>
+#include <cholmod.h>
 #else
 
 // constants
--- a/library/src/cmake/suitesparse/FindCHOLMOD.cmake
+++ b/library/src/cmake/suitesparse/FindCHOLMOD.cmake
@@ -1,5 +1,5 @@
 # ########################################################################
-# Copyright (C) 2024 Advanced Micro Devices, Inc. All rights reserved.
+# Copyright (C) 2024-2025 Advanced Micro Devices, Inc. All rights reserved.
 #
 # Permission is hereby granted, free of charge, to any person obtaining a copy
 # of this software and associated documentation files (the "Software"), to deal
@@ -45,7 +45,10 @@ This module defines the following variables:
 
 #]=======================================================================]
 
-find_path(CHOLMOD_INCLUDE_DIR suitesparse/cholmod.h)
+find_path(CHOLMOD_INCLUDE_DIR
+    NAMES cholmod.h
+    PATH_SUFFIXES suitesparse
+    )
 find_library(CHOLMOD_LIBRARY cholmod)
 
 include(FindPackageHandleStandardArgs)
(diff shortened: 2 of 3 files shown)


jmachado-amd:
Hi @stellaraccident, thanks! I'll have a look at your PR tomorrow, in the meantime, please target the fix at the `develop` branch.

jmachado-amd (commented):
Hi @stellaraccident, it looks like my week was far busier than anticipated -- apologies for the delay. I see two other changes that are required in order to allow `hipSOLVER` to find its Suitesparse dependencies in a general setting:

https://github.com/ROCm/hipSOLVER/blob/d3948089123cf30ed67e4208237c533f23e4095f/library/src/cmake/suitesparse/FindCHOLMOD.cmake#L48
should be updated into:
```.cmake
find_path(CHOLMOD_INCLUDE_DIR
    NAMES cholmod.h
    PATH_SUFFIXES suitesparse
    )
```

https://github.com/ROCm/hipSOLVER/blob/d3948089123cf30ed67e4208237c533f23e4095f/library/src/cmake/suitesparse/FindSuiteSparse_config.cmake#L48
should be updated into:
```.cmake
find_path(SUITESPARSE_CONFIG_INCLUDE_DIR
    NAMES SuiteSparse_config.h
    PATH_SUFFIXES suitesparse
    )
```
Thanks again for taking the time to make the PR, and please let me know if this works for you.

cgmb (commented):
If this is wrong, then mea culpa. I'd hoped that Spack would catch this sort of error, but I guess they had a patch for an earlier problem which ended up sidestepping this issue. :(

The bit that confuses me is that the `SuiteSparse::CHOLMOD` target is actually quite a new thing in the upstream SuiteSparse project. It didn't exist in older versions, so it's weird that there's an expectation that the user will add an include path to their build flags just to use the OS-provided SuiteSparse headers.

stellaraccident:
Yeah, it's kind of a head scratcher if trying to unwind how this got this way and ever worked right. The details on this are context switched out of my head, but I have an easy repro to try changes with a local built SuiteSparse.

jmachado-amd:
@stellaraccident, I've opened a pr with the changes I suggested last week. They work in my own tests, but please let me know if you can try them in your reproducer.

jmachado-amd:
Hello @renjithravindrankannath, we would be very grateful if you could test this PR in spack. Many thanks in advance!

renjithravindrankannath (approved):
Verified the change in Spack mainline. Looks good to me.

jmachado-amd (approved):
This is all that I need approve the PR, at this point I would just move to rebase and merge and start a cherry-pick into 6.4.

@cgmb, do you have anything to add?

tfalders (approved):

cgmb (approved):
I still think the suitesparse behaviour is a bit weird, but I see that with the find module update provided by Julio, this change builds properly with libsuitesparse-dev 1:5.10.1+dfsg-4build1 from Ubuntu 22.04 (which did not include CHOLMODConfig.cmake). Thanks @stellaraccident and @jmachado-amd.

The User Guide for CHOLMOD also includes a sample program that shows the direct inclusion of `cholmod.h`.

PR #368, opened 2025-02-24
SWDEV-516659 - Ensure the list of recommended and suggested packages are comma separated

raramakr:
A list in cmake is a semicolon(;) separated group of strings. The cpack recommends and suggests variables should have comma separated list, otherwise it will result in package installation failure.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -250,7 +250,10 @@ if(NOT USE_CUDA)
     rocm_package_add_dependencies(DEPENDS ${hipsolver_pkgdeps})
   else()
     list(APPEND CPACK_DEBIAN_RUNTIME_PACKAGE_RECOMMENDS "rocsparse" ${hipsolver_pkgdeps})
+    # Cpack Recommends/Suggests list should be comma separated
+    string(REPLACE ";" "," CPACK_DEBIAN_RUNTIME_PACKAGE_RECOMMENDS "${CPACK_DEBIAN_RUNTIME_PACKAGE_RECOMMENDS}")
     list(APPEND CPACK_RPM_RUNTIME_PACKAGE_SUGGESTS "rocsparse" ${hipsolver_pkgdeps})
+    string(REPLACE ";" "," CPACK_RPM_RUNTIME_PACKAGE_SUGGESTS "${CPACK_RPM_RUNTIME_PACKAGE_SUGGESTS}")
   endif()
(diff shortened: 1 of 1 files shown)


pramenku:
We need to CP to 6.4 also

qjojo:
I may be stepping on a land mine by asking this, but CMake 3.12 added a more idiomatic way to do this https://cmake.org/cmake/help/latest/command/list.html#join

`list(JOIN CPACK_DEBIAN_RUNTIME_PACKAGE_RECOMMENDS "," CPACK_DEBIAN_RUNTIME_PACKAGE_RECOMMENDS)`

We would have to bump the minimum version to use it, but I can see that rocBLAS targets 3.16, and I believe the cmake version shipped with the OSes we support is much higher than that.

raramakr:
> I may be stepping on a land mine by asking this, but CMake 3.12 added a more idiomatic way to do this https://cmake.org/cmake/help/latest/command/list.html#join
> 
> `list(JOIN CPACK_DEBIAN_RUNTIME_PACKAGE_RECOMMENDS "," CPACK_DEBIAN_RUNTIME_PACKAGE_RECOMMENDS)`
> 
> We would have to bump the minimum version to use it, but I can see that rocBLAS targets 3.16, and I believe the cmake version shipped with the OSes we support is much higher than that.

Since the current cmake version is set to 3.7 didn't use cmake list(JOIN ..) in this patch. Also this need to be cherry-picked to 6.4 release branch. 
Maybe we can bump the cmake version and use cmake  list(join ... ) in another patch

cgmb (approved):
IMO, you might as well just directly set the `*_RECOMMENDS` and `*_SUGGESTS` with the `string` command, rather than using `list` and then `string`. With that said, I can live with it as-is if you just want to merge something that you've tested ASAP. It's not like the rest of our CMake is beautiful anyway.

#364 is related and is also desired for a cherry-pick, once reviewed and approved.

cgmb:
> Maybe we can bump the cmake version and use cmake list(join ... ) in another patch

Sure. Bumping the minimum requirement for hipSOLVER to CMake 3.12 is fine. The rocSOLVER library requires 3.13, and rocBLAS requires 3.16.

pramenku:
Can we proceed to merge this PR and put in "staging" branch for further CQE cycle?

tfalders (approved):

PR #369, opened 2025-02-25
SWDEV-516659 - Ensure the list of recommended and suggested packages are comma separated

raramakr:
A list in cmake is a semicolon(;) separated group of strings. The cpack recommends and suggests variables should have comma separated list, otherwise it will result in package installation failure.


jmachado-amd (approved):
Approving as is, given that #368 has already been approved. Thanks @raramakr!

tfalders (approved):

PR #374, opened 2025-03-01
Cherry pick CI static analysis fix

samjwu:
RE: requires cherry pick approval before merge


tfalders:
@samjwu It looks like PR #400 has included the same changes for cherry-pick to 6.4. Can we close this PR since it is now redundant?

PR #385, opened 2025-04-02
Re-enable `sygvdx` and `hegvdx` tests

jmachado-amd:


ammallya:

PR #392, opened 2025-04-16
[DRAFT] Enable complex functions for csrlsvqr

qjojo:
Local testing showed these functions as working again, this PR is to verify they're working on all our supported architectures.

### DIFF
--- a/clients/include/hipsolverSp.hpp
+++ b/clients/include/hipsolverSp.hpp
@@ -1,5 +1,5 @@
 /* ************************************************************************
- * Copyright (C) 2023-2024 Advanced Micro Devices, Inc. All rights reserved.
+ * Copyright (C) 2023-2025 Advanced Micro Devices, Inc. All rights reserved.
  *
  * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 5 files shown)


clients/include/hipsolverSp.hpp
```diff
@@ -308,5 +308,5 @@ inline hipsolverStatus_t hipsolver_csrlsvqr(bool                          HOST,
                                     singularity);
     else
         return HIPSOLVER_STATUS_NOT_SUPPORTED;
-}*/
+}
```
EdDAzevedo, line 311: If we'll be just copying the sparse matrix to fully dense matrix and call a dense solver, perhaps we can do this on the host as well to call lapack and rocsolver on GPU. I think it should be straight-forward to write C++ code to convert CSR to fully dense matrix on cpu. Perhaps AI copilot can generate most of the code.

EdDAzevedo (approved):
minor suggestion to consider implementing both the gpu and cpu host versions using the same approach of converting CSR sparse matrix to fully dense matrix and use lapack or rocsolver to perform the computations.

PR #397, opened 2025-05-08
Check if rocblas_DIR env var is set before using it

zichguan-amd:
Check if `rocblas_DIR` env var is set before using it on Windows

TheRock: https://github.com/ROCm/TheRock/issues/527


tfalders (approved):

cgmb (approved):
We really should just get rid of the uses of that variable.

PR #398, opened 2025-05-15
Use HIPSOLVER_BUFFERSIZE_RETURN_BYTES to restore original bufferSize behavior

tfalders:


amd-jnovotny (approved):
New changelog note looks okay, but I noticed the changelog has 6.5 release notes even though 6.5 has been cancelled. The notes in this section should be moved to 7.0 when you have the opportunity (doesn't necessarily need to be done as part of this PR).

jmachado-amd (approved):
It looks good to me, many thanks @tfalders!

jmachado-amd:
> New changelog note looks okay, but I noticed the changelog has 6.5 release notes even though 6.5 has been cancelled. The notes in this section should be moved to 7.0 when you have the opportunity (doesn't necessarily need to be done as part of this PR).

On a tangential note: As some of those features may appear in 6.4, let's also try to move the entry that notifies users about the upcoming changes to buffer sizes from the 6.5 changelog into 6.4.x.

cgmb (commented):
I'm not so sure about a global runtime setting that fixes a problem that is presumably a local compile-time issue... I'm a bit worried that this might ultimately cause as many problems as it fixes.

If you're sure about this, then it's certainly not my place to tell you what to do. It may be worth giving the design one more discussion, though.

tfalders:
> I'm not so sure about a global runtime setting that fixes a problem that is presumably a local compile-time issue... I'm a bit worried that this might ultimately cause as many problems as it fixes.
> 
> If you're sure about this, then it's certainly not my place to tell you what to do. It may be worth giving the design one more discussion, though.

I considered adding an API that would allow users to enable the old behaviour at runtime, but I don't have anywhere to store the result of that API since the hipSOLVER handle is just an alias of the rocBLAS handle. That could be changed, but I have a slight worry that someone out there might be assuming this is the case, kind of like how we're using hipSPARSE components in hipSOLVER.

AGonzales-amd (approved):

PR #422, opened 2025-09-11
Fix dependency / race condition

marbre:
## Motivation

Allow to compile hipSOLVER via TheRock (https://github.com/ROCm/TheRock) in the first run.

## Technical Details

Compiling the `hipsolver_fortran_client` target fails as it is not guaranteed that `hipsolver.mod` is created in time during the first build attempt. A second run is successful though. Using `target_link_libraries` resolves this.

https://discourse.cmake.org/t/cmake-3-27-8-regression-with-ninja-and-fortran-dependencies/9920/4

## Test Plan

N/A

## Test Result

N/A

## Submission Checklist

- [x] Look over the contributing guidelines at https://github.com/ROCm/ROCm/blob/develop/CONTRIBUTING.md#pull-requests.


stellaraccident (approved):
I have researched this issue and spent time tracing through what is going on. The theory I expressed many months ago was correct, and your solution is the most correct way to do this.

For those interested, see the discussion where Brad King says the same thing : https://discourse.cmake.org/t/cmake-3-27-8-regression-with-ninja-and-fortran-dependencies/9920/5.

The issue is that .mod files are like generated headers in the C world, produced as a byproduct of a previous library step. Adding a target level dependency merely specifies an ordering such that in this case parts of the producer and consumer commands can execute in parallel. The CMake ninja generator uses file level granularity to enable greater concurrency and will allow these to execute at the same time (so that the producer writes the mod file last and the first thing the consumer tries to do is read it). There are a couple of ways to get the file level dependencies right, but the canonical one is target_link_libraries: that adds rules for dynamic module scanning of the deps and adds dynamic file level dependencies based on what it finds in a similar way to how C/C++ generated header scanning gets hooked up into the build graph (slightly different but same mechanic). This is the proper way to express a fortran library dependency.

tfalders (approved):

jmachado-amd (approved):

ammallya:

marbre:
> Imported to ROCm/rocm-libraries

@ammallya Why wasn't this merged instead of being exported?

PR #426, opened 2025-09-18
Fix syntax error in clients/gtest/CMakeLists.txt

ScottTodd:
See https://cmake.org/cmake/help/latest/variable/ENV.html

Follow-up to https://github.com/ROCm/hipSOLVER/pull/397


tfalders (approved):

jmachado-amd (approved):

ammallya:

